PR #8, opened 2013-06-19
Add guideline to use map/select instead of each.

clizzin:


tommydangerous:
How come this never got merged in?

jasonkb:
@tommydangerous what ericlevine said: "Not sure how I feel about codifying worse-performing code into our style guide when it probably should be determined on a case-by-case basis. My instinct is to :-1:"

PR #14, opened 2013-07-02
Add rules for multi-line method calls & hash literals

clizzin:
I'd like if we did this consistency so it's faster to grok new code.

@robotpistol since we have differing opinions.


tommydangerous:
PR still needed?

tommydangerous:
We will have a rule similar to this in the coming weeks, for now I will close.

PR #19, opened 2013-07-17
Ban ActiveRecord dynamic finder methods

clizzin:
I was reminded by @gkoo in [this comment](https://github.com/airbnb/airbnb/pull/3114/files#r5206241) that dynamic finder methods are deprecated.  Moreover, they are lame.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -977,6 +977,16 @@ in inheritance.
 
 ## Rails
 
+  * Do not use ActiveRecord dynamic finders.  They are [deprecated](http://guides.rubyonrails.org/active_record_querying.html#dynamic-finders).
+
+    ```ruby
+    # bad
+    Parties.find_by_country_and_wildness('US', 9000)
+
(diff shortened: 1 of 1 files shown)


BMorearty:
There's been some confusion about deprecation of ActiveRecord dynamic finder methods. In particular, [`find_by_` is not deprecated](https://github.com/rails/activerecord-deprecated_finders). I'd rather not add this guideline.

tommydangerous:
Newer versions of Rails does:
```
User.find_by(id: 1)

Plane.find_by(wing_span: 10)
```

They are very handy.

tommydangerous:
I think we should close this since the proper way to do this in Rails these days is to do:
```
User.find_by(name: "Ragna")
```

And since we are on Rails 3, we can't do this. I don't think we should promote a style that we can't accomplish at the moment.

PR #21, opened 2013-08-13
Add guideline to wrap ActiveRecord model scopes in lambdas

clizzin:
Let us enshrine this knowledge forever.

cc @schleyfox


tommydangerous:
Lambdas over Procs in scopes. However, we should encourage

```
def self.active
  ...
end
```

over

```
scope :active ...
```

tommydangerous:
^ Agree with the above. Lambdas over Procs. But we should encourage:

```
def self.active
  ...
end
```

over

```
scope :active, ...
```

BMorearty:
Why not use scopes, @tommydangerous? There's nothing wrong with them and they're handy. 

P.S. Personally, I would change the syntax in the example to a stabby proc but I don't think there's a need to enshrine that in a guideline about one vs. the other.

robotpistol:
:+1:  for choosing one type of lambda and running with it.

README.md
```diff
@@ -1006,6 +1006,18 @@ in inheritance.
     end
     ```
 
+* When defining ActiveRecord model scopes, wrap the relation in a `lambda`.  A
+  naked relation forces a database connection to be established at class load
+  time (instance startup).
+
+    ```ruby
+    # bad
+    scope :foo, where(:bar => 1)
```
romanfuchs: I assume this is the case for this syntax as well:
```rb
scope :active, :conditions => ["deleted_at IS NULL"]
```
Let's maybe add this example as well, since we have it a lot in the codebase.

romanfuchs:
Quite a bit of cleanup left for us then: https://dashboards.d.musta.ch/codesearch?q=scope%5C+%5C%3A%5B%5E%5C%7B%5D*%24&f=&case_sensitive=on

README.md
robotpistol: that can also be represneted as 
``ruby
scope :active, where(:deleted_at => nil)
```

robotpistol: @romanfuchs I'm guessing the majority of those like you have in your example are older.  Most newer code in our codbase prefers the use of `where` blocks to `:conditions`

tommydangerous:
I am going to update this PR using lambda literals

tommydangerous:
Okay, can we get a LGTM or :+1: ?

BMorearty:
@tommydangerous The stabby lambda syntax is good. But you also changed the hash syntax from `:bar => 1` to `bar: 1`. Let's separate that out to a separate PR, since all existing hashes in this guide use `:bar => 1`. (Personally I would prefer having a separate PR that says to use the older `:bar => 1` syntax, because the newer syntax only works with symbols as keys, so with the newer syntax you end up by necessity having both syntaxes throughout the code.)

tommydangerous:
K, changing it.

venantius:
Sure, :+1:

robotpistol:
:+1:

venantius:
rebase good sir

README.md
robotpistol: we should probably have another PR to address prefered lambda syntax

BMorearty: > we should probably have another PR to address prefered lambda syntax

That makes sense.

PR #22, opened 2013-09-05
Curly braces should not have spaces around them

horaceko:
From the style guide:

* Use spaces around operators; after commas, colons, and semicolons; **and around `{` and before `}`**.

One of these things is *not* like the others. 

To me, `{` and `}` are closer in function to `(`, `[`, `)`, and `]` than commas, colons, and semicolons. As such, I don't feel that we should treat them differently in terms of whitespace padding around those symbols. Here's what I think the docs *should* say:

* Use spaces around operators; after commas, colons, and semicolons.
* No spaces after `(`, `[`, `{` or before `}`, `]`, `)`.

Isn't that so much more consistent?

/cc @robotpistol and anyone else who may want to jump in on this inevitable nerd fight.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,21 +84,20 @@ We also maintain a [JavaScript Style Guide][airbnb-javascript].
 
 * Never leave trailing whitespace.
 
-* Use spaces around operators; after commas, colons, and semicolons; and around
-  `{` and before `}`.
+* Use spaces around operators; after commas, colons, and semicolons.
 
     ```Ruby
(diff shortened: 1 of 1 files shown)


tommydangerous:
This PR is no longer needed right?

Can we please not do `{|e| e + 2}`. I strongly prefer `{ |e| e + 2 }`

robotpistol:
I think this can safely be closed since the last activity was 2013

PR #31, opened 2014-09-22
Change the suggestion for public, protected, and private indentation.

nathanielweinman:
I think that indenting these keywords one space less than the functions
they apply to leads to much cleaner code. public, protected, and private
are effectively marking off a block of code, so visually they should not
visually blend in with the code affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,8 +856,9 @@ in inheritance.
     end
     ```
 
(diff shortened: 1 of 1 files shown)


tommydangerous:
PR no longer needed?

PR #38, opened 2015-02-11
Syntax highlight two code examples

maxjacobson:
Kind of a strange fix, yes, but it gets rid of this

    ```ruby


clizzin:
Thanks!

PR #39, opened 2015-02-20
Corrected parameter & argument misnomers

sethkrasnianski:
A common misnomer in programming is the use of the words argument and parameter. When writing a function or method's signature (declaring the method), the values being passed to a method are called `parameters`, not `arguments`. When calling a function or method, the values passed to the call are formally known as `arguments`.

Example of parameters:
```ruby
def method_with_parameters(param1, param2)
    param1 + param2
end
```
Example of arguments:
```ruby
method_with_parameters(1, 2) # 1 & 2 are arguments
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
            end
     ```
 
-* Align function arguments either all on the same line or one per line.
+* Align function parameters either all on the same line or one per line.
 
     ```ruby
     # good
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -398,20 +398,20 @@ Never leave commented-out code in our codebase.
...
      end
 
-     def some_method_with_arguments(arg1, arg2)
+     def some_method_with_parameters(arg1, arg2)
        # body omitted
      end
      ```
 
-* Do not use default arguments. Use an options hash instead.
+* Do not use default parameters. Use an options hash instead.
```
clizzin: In this case, we really do mean to use "arguments," not "parameters."

clizzin:
Thanks for the contribution!  I'm not personally strict about proper usage of these terms since it's easy to understand what people mean even if they use the wrong term, but I'm happy to accept this pull request and make the style guide use the terms properly.

My only nit is that "default arguments" really does mean "default arguments" — we define default values/arguments for the parameters.  If you push up an additional commit to fix that, I'll go ahead and merge!

README.md
sethkrasnianski: Ahh yes, the hash _inside_ the method is what you're referring to, **Not** `options = {}`?

BMorearty:
I feel the same way as @clizzin -- I'm not strict with the proper usage of these terms but am ok with the change. Agree with 2nd paragraph too.

sethkrasnianski:
Updated the nit @clizzin

clizzin:
Great, thanks!

PR #40, opened 2015-03-12
"Omit parentheses" should be at the same level as "Use parentheses"

aifreedom:
to: @clizzin 

As the title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ Never leave commented-out code in our codebase.
     f(3 + 2) + 1
     ```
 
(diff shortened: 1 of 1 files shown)


clizzin:
Good catch.  Can we add an additional revision to make it look like this:

```
**Omit parentheses** for a method call:

* If the method accepts no arguments:
```

That will make the format analogous the "Use parentheses" section.

Thanks for the fix!

tommydangerous:
Has the change been made already or is this PR no longer needed?

robotpistol:
Loks like it's already in

PR #41, opened 2015-03-20
Fix "bad" version of do-end example.

BMorearty:
Based on what was said in the previous paragraph,
I believe this example was meant to show that we
don't want "do...end" on a single line.


clizzin:
:+1: Thanks!

PR #42, opened 2015-03-24
Add indentation rule for multi-line boolean expression

siclait:
I believe this is easier to read and is worth pointing out for individuals who are coming from a non-Ruby background where this situation does not arise (because there may not be standalone boolean expressions)

to: @clizzin

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,24 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
     end
     ```
 
+* Indent succeeding lines in multi-line boolean expressions.
+
+    ```ruby
+    # good
+    def is_eligible?(user)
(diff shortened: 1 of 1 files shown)


spikebrehm:
LGTM

BMorearty:
LGTM

robotpistol:
LGTM

robotpistol:
Arguable we might want to write sure a method in a different way like:
```ruby
def is_eligible?(user)
  return false unless Trebuchet.current.launch?(ProgramEligibilityHelper::PROGRAM_TREBUCHET_FLAG)
  return false unless is_in_program?(user)
  return false unless program_not_expired
  true
end
```

But that style should not be mandated.

BMorearty:
Agreed.

clizzin:
Sounds like we have consensus that this guideline is good, so merging.  Thanks for contributing!

PR #43, opened 2015-03-25
fix a condition style

garygangwu:
Always prefer a simplified condition statement

to: @clizzin

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,7 +568,7 @@ In either case:
       end
 
       # okay
-      if !(foo? && bar?)
+      if !foo? || !bar?
         ...
       end
     ```


BMorearty:
Both are okay. (`if !(foo? && bar?)` and `if !foo? || !bar?`). Which one is clearer? It depends on the situation. So I wouldn't change this example, because changing it muddles the recommendation, which is about whether to use `unless` on a negative condition.

garygangwu:
As far as I know, many language convention prefer a simplified form of compound conditional statements, such as

`A && B && C`
`A || B || C`

in which `A` can either be a `!condition` or `condition`

these complex compound conditional statements (with `( )`) are highly discouraged, due to the readability and maintenance concern.

`!(A && B && C)`
`A && (B || C)`

I think ruby language shouldn't be an exception here, but I totally understand this is a very personal preference.

andykram:
Agreed with @BMorearty

schleyfox:
I think it's easier to read a single negation with an AND than it is to read a double negation with OR.  The use of parens doesn't make it complex, it just allows us to negate the simplest condition.

robotpistol:
Agreed. I don't think this is something that needs mandating in the style guid.

clizzin:
I agree with Brian that the right choice between `!(foo? && bar?)` and `!foo? || !bar?` is highly dependent on the context of the code, and what the exact method names are.  In some cases, the first form will be more readable, and in others, the second form will be more readable.  So I don't think a style guide should state a preference either way, and it seems like there's enough people who agree that we should close this pull request.

Thanks for making the suggestion though!

PR #44, opened 2015-03-26
Make an explicit return in Ruby

garygangwu:
This may be a very controversial topic. Many people love the Ruby way that the `return` keyword is completely unnecessary. However, as we scale our engineering team, we may have to seriously consider to limit some of the Ruby freedom, so that we can minimize the chances of introducing software bugs.

@jasonkb @clizzin @BMorearty 

anyone else would like to be cc-ed

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,18 +677,98 @@ In either case:
     `{...}`, but they should ask themselves if this code is really readable and
     whether the block's content can be extracted into nifty methods.
 
-* Avoid `return` where not required.
+* Highly recommend to use `return` even if it is not required in Ruby language.
+
+    Although it is a Ruby way that an explicit return is unnecessary, it may introduce
+    confusions to programmers working with unfamilar code and to programmers learning
+    Ruby from other strict languages.
 
     ```Ruby
-    # bad
+    # bad: it is unclear whether this function want to return the updated value
+    #      or just want to set the new value
+    def update(new_value)
+      @value = func(new_value)
+    end
+
+    # Someone may extend the function by add a log message, and then unintentionally
+    # complete break it if the return value is intended
+    def update(new_value)
+      @value = func(new_value)
+      Rails.logger.info '...'
+    end
+
+    # good: make it clear the return value is intended
+    def update(new_value)
+      return (@value = func(new_value))
+    end
+    ```
+
+    ```Ruby
+    # bad: it is unclear whether it just wants to invoke another method, or returns
+    #      the function result
     def some_method(some_arr)
-      return some_arr.size
+      if condition
+        func_1(some_arr)
+      else
+        func_2(some_arr)
+      end
     end
 
     # good
     def some_method(some_arr)
-      some_arr.size
+      if condition
+        return func_1(some_arr)
+      end
+      return func_2(some_arr)
     end
+    ```
+
+    ```Ruby
+    # good: just have "return" here to make 100% clear to any programmer
+    def some_method
+      ...
+      return value
+    end
+
+    # good, but why bother, adding "return" incurs no cost
+    def some_method
+      ...
+      value
+    end
+    ```
+
+    ```Ruby
+    # good: no "return" in a function with one statement
+    def default_options
+      {
+        :key => '...',
+        :value => 100
+      }
+    end
+    ```
+
+* Avoid "return" under these situations
+
+    ```Ruby
+    # bad: no "return" in a do-end yield block
+    def foo
+      ...
+      array.map do |v|
+        return v.upcase
+      end
+      ...
+    end
+    ```
+
+    ```Ruby
+    # bad: Proc.new will extend the block of the current function, instead
+    #      of becoming a sub function
+    def foo
+      Proc.new { return "inside the proc" }.call
+      return "end of foo"
+    end
+
+    puts foo # "inside the proc" to be printed
     ```
 
 * Using the return value of `=` (an assignment) is ok, but surround the


README.md
```diff
@@ -677,17 +677,64 @@ In either case:
...
+
+    Although it is a Ruby way that an explicit return is unnecessary, it introduces
+    confusions to programmers working with unfamilar code and to programmers learning
+    Ruby from other strict languages.
 
     ```Ruby
-    # bad
+    # bad: it is unclear whether this function want to return the updated value
+    # or just want to set the new value
+    def update(new_value)
```
bf4: Or

```ruby
# @param new_value [Type] the new value
# @return updated value [OtherType]
```

garygangwu: any comment to describe the input and output is highly desired

andykram: I would argue that the amendment here should be something along the lines of 'when you don't wish to return the last value from a method, return `nil` instead'. That way it's clear when someone doesn't desire to return the value of an assignment.

README.md
```diff
@@ -677,20 +677,101 @@ In either case:
...
+    # complete break it if the return value is intended
+    def update(new_value)
+      @value = func(new_value]
+      Rails.logger.info '...'
+    end
+
+    # good: make it clear the return value is intended
+    def update(new_value)
+      @value = func(new_value]
+      return @value
```
thisnick: `return (@value = func(new_value))`?

README.md
```diff
@@ -677,20 +677,101 @@ In either case:
...
+    Although it is a Ruby way that an explicit return is unnecessary, it introduces
+    confusions to programmers working with unfamilar code and to programmers learning
+    Ruby from other strict languages.
 
     ```Ruby
-    # bad
+    # bad: it is unclear whether this function want to return the updated value
+    #      or just want to set the new value
+    def update(new_value)
+      @value = func(new_value]
```
thisnick: close with `)` and not `]`.

README.md
```diff
@@ -677,20 +677,101 @@ In either case:
...
+
+* Avoid "return" under these situations
+
+    ```Ruby
+    # bad: no "return" in a do-end yield block
+    def foo
+      puts yield
+    end
+
+    foo do
```
thisnick: this example is a little contrived. Maybe just
```
array.map do |v|
  return v.upcase
end
```

README.md
boyers: You could also take out the parentheses. But I think it's best the way @garygangwu wrote it.

thisnick: I think there are too many lines for a one line function.

README.md
garygangwu: good point

README.md
garygangwu: nice catch!

README.md
garygangwu: yeah, it can be simplified. Thanks for the tip

README.md
```diff
@@ -677,20 +677,101 @@ In either case:
...
+    # Someone may extend the function by add a log message, and then unintentionally
+    # complete break it if the return value is intended
+    def update(new_value)
+      @value = func(new_value]
+      Rails.logger.info '...'
+    end
+
+    # good: make it clear the return value is intended
+    def update(new_value)
+      @value = func(new_value]
```
robotpistol: close with `)` and not `]`. here too

garygangwu:
thanks for the feedbacks, updated the PR

LogicWolfe:
I am opposed to this change, here are a few reasons:

1. It's inconsistent with our current codebase and the generally agreed best practices to date, which will add a fairly significant switching cost.
2. I don't think it improves clarity. The idea that you only need to worry about the return value in the case where `return` is used explicitly is a trap. There is _always_ a return value in Ruby and it _always_ is important to be aware of it.
3. I don't think this is an important enough subject to be enshrined in our style guide. The style guide is most useful when it is most concise. The more we add to it that is of questionable utility, the less people will be able to internalize the guide and apply it consistently. Having consistent style is important, but knowing where to draw the line is also important, I don't think this meets the bar.

I have no objection to people using `return` statements where they feel it improves the clarity of their intent, but I don't think we should enforce it in all cases.

robotpistol:
I also disagree with this change, for the same reasons as @LogicWolfe

jasonkb:
Also want to point out that it's self-inconsistent.

```
   # bad: it is unclear whether this function want to return the updated value
   #      or just want to set the new value
   def update(new_value)
     @value = func(new_value)
   end
```

but later

```
   # good: no "return" in a function with one statement
   def default_options
     {
       :key => '...',
       :value => 100
     }
     end
```

jasonkb:
I also disagree with this change, for the same reasons as @LogicWolfe.

jasonkb:
The change also doesn't way, what is the "good" alternative to this?

```
   # bad: it is unclear whether this function want to return the updated value
   #      or just want to set the new value
   def update(new_value)
     @value = func(new_value)
   end
```

Is it 

```
   # bad: it is unclear whether this function want to return the updated value
   #      or just want to set the new value
   def update(new_value)
     @value = func(new_value)
      nil
   end
```

?

I think the only self-consistent guideline that would work is "Every method has an explicit return at the bottom". But this is so draconian and ugly and non-Ruby. And we'd have to add thousands of lines of return statements to our codebase to make our code conformant.

romanfuchs:
+1 to @LogicWolfe

schleyfox:
+1 to @LogicWolfe

garygangwu:
Thanks for the thoughtful feedback (+1 to @LogicWolfe too). It is very valuable to learn from your prospective. Certainly, as mentioned in the PR description, it is a very controversial topic. I totally understand that no `return` is an unique feature that makes it differentiated from other programming languages. I am also certain that I will be killed immediately if I make a proposal to the python or C++ community to deprecate `return` keyword.

Do you guys feel the current standard is also too `narrow minded`? 

```
Avoid return where not required.

# bad
def some_method(some_arr)
  return some_arr.size
end

# good
def some_method(some_arr)
  some_arr.size
end
```

README.md
```diff
@@ -677,20 +677,100 @@ In either case:
...
+    ```Ruby
+    # bad: Proc.new will extend the block of the current function, instead
+    #      of becoming a sub function
+    def foo
+      Proc.new { return "inside the proc" }.call
+      return "end of foo"
+    end
+
+    puts foo # "inside the proc" to be printed
+    ```
```
clizzin: Although people generally don't like the advice to explicitly `return` values at the end of a method, I do think this advice on `yield` vs. `return` is useful.  New Ruby developers often misunderstand what it means to `return` from a block or proc, so I think it would be helpful for the style guide to clarify this behavior.  Can you open a new pull request with just this bullet point?

clizzin:
Seems like there's consensus that we shouldn't require explicit `return` statements.  I am of this mind as well.

Gary's latest comment raises a new question though, which is whether we should accept both implicit and explicit return values, and update the style guide to not mention a preference on this at all.  What do people think?

robotpistol:
I think in Ruby's Case because of the differences in how returning from blocks actually works preferring a lack of `return` is probably actually safer than the alternative.

BMorearty:
Thanks for the suggestion. I disagree with it.

@garygangwu wrote:
> it is a very controversial topic

I don't think it's a controversial topic in Ruby-land. This is the first time I've seen anyone prefer the use of the optional `return` statement at the end of a method in Ruby.

> Do you guys feel the current standard is also too narrow minded? "Avoid return where not required."

I don't feel it is. I agree with the current standard.

clizzin:
Okay, seems like there's enough consensus on sticking to the current guideline that we should stick with it.  Thanks for participating in the discussion, everyone.

Gary, I can see how implicit returning/yielding of values is confusing to those coming from outside Ruby.  But in the Ruby ecosystem, it's pretty normal, and as Allen and Nathan pointed out, it reflects the way the language actually works under the hood — the last value is *always* yielded or returned.

panke6670:
hope it's not too late, i think given the case that many engineers are already familiar with current ruby coding style, it doesn't make too much sense to enforce "return". however, I also feel we shouldn't discourage engineers to write code with explicitly return (which is the case in current coding style guide). and we may have a slightly different coding style in our code base (which is the downside), but it does make certain code easier to understand (for example, in a long function, i do want to write return explicitly to remind myself that code ended there).
so long story short, can we not discourage ppl from using return(even when not needed?).

LogicWolfe:
@panke6670 Maybe open a new PR with that change so we have a clean place to discuss it? I would be in favor of leaving it to the discretion of the implementer.

panke6670:
will do

garygangwu:
@BMorearty , really appreciated of the direct feedback, and I highly respect your ruby experiences. 

> I don't think it's a controversial topic in Ruby-land. This is the first time I've seen anyone prefer the use of the optional return statement at the end of a method in Ruby.

I consider myself as a junior ruby coder (but probably a mature C++ coder). In the ruby world, I think the `return` discussion is not a rare discussion. Here are some of debates easily found online,

http://stackoverflow.com/questions/1023146/is-it-good-style-to-explicitly-return-in-ruby
https://github.com/bbatsov/ruby-style-guide/issues/9
https://github.com/bbatsov/ruby-style-guide/pull/150

Frankly, a lot of programmers similar to me who constantly switching languages does feel that it is uneasy to have a function without explicit `return`.  I totally recognize that `return` is a great syntactic sugar of Ruby for simplicity, but I highly disagree that we should discourage `return` when it is unnecessary. I think we should also appreciate the programmers who makes explicit `return`

schleyfox:
Wow, I'm the second answer in that stack overflow post (from 6 years ago).

Some languages use explicit returns, others do not.  Ruby happens to be a language that does not favor explicit returns, and to ruby developers an explicit return looks like something fishy is going on.  Explicit returns actually cause you to look for the reason that it's there, because it's a break from the style.  When using Ruby, we should follow the style of Ruby.

Ruby has a lot of influence from functional languages that eschew explicit returns (Lisp, Erlang, and ML, at the very least) because it emphasizes that void functions indicate impurity, and functions should produce values.  I think it's also telling that the basic lambda form in java 8 is an implicit return.

README.md
bf4: `return, break, next` do different things in blocks than you might expect. :shrug:

schleyfox:
The last point I'll make is that this style guide is an attempt to encourage a good ruby style that is also generally compatible with the prevailing style of the ruby community.  If we come up with a style that varies from what is practiced, we run into a situation where our internal ruby differs from that of the open source libraries that we depend on.  If this happens, digging into library code to figure out how things work becomes a more jarring, and therefore difficult, endeavor.

garygangwu:
@schleyfox, thanks for the explanation. Wow, so impressive that your 6 year old post got highlighted here again. Likely we will dig out more and more your past posts, and make it easier for you write a autobiography。

This is a good point that our coding standard should't be so different from open source convention. Totally agree that our coding style shouldn't be very different from the open source community. On the other side, when google introduce its own C++ coding standard, it is slightly different from some C++ code bases outside. But I haven't heard that google engineers run into a big trouble. Of course, again i totally agree that our engineers should be able to easily dig into any gem written by external engineers. That is the basic survival skill. 

One last question, if one of our programmers made an explicit `return` in the last line of the function, shall we ask her or him to fix it otherwise reject the PR, or leave as it is?

robotpistol:
I think it may be suggested but I probably wouldn't interpret it as a blocker.  It will probably be expressed as "This might look cleaner without the return" or "You don't need the return"

fengmingAirbnb:
Personally, I found the return statement sometimes useful and the strong enforcement of the `no-return` rule not well grounded. For instance, I have seen many times in our code base things like:
```
def foo
  result = nil
  if condition_A
    result = sth_A
  else
    ...
    #{A long tedious block with 30 ~ 40 lines}
    ...
  end
  result
end
```

I think the code would read slightly better if we do:
```
def foo
  if condition_A
    return result
  ...
  #{A long tedious block with 30 ~ 40 lines that I don't really care}
  ...
end
```

Even worse, I have seen the following in our code base:
```
def bar
  result = nil

  if result == nil and condition_A
    result = result_A
  end

  if result == nil and condition_B
    result = result _B
  end

  #{So many conditions that I don't understand}

  if result == nil and condition_Z
    result = result_Z
  end

  if condition_evil
    # Some evil or careless programmer might forgot to check result == nil and change result here.  
  end

  result
end
```

I think the code would look cleaner if we do:
```
def bar
  if condition_A
    return result_A
  end

  if condition_B
    return result _B
  end

  #{So many conditions that I don't understand and don't care}
end
```

IMHO, in these cases, returning early can reduce both the unnecessary indentation and human errors.
Probably my understanding of Ruby is very limited and there are other ways to improve the above code.

clizzin:
@fengmingAirbnb I think you misunderstand the guideline.  The guideline does not advise against use of `return` to return a value early in the method body, but rather against the use of `return` on the last line of a method/proc.  I personally prefer the examples you give of using early returns to make code read more smoothly, and I bet many others on this thread would as well.

BMorearty:
@fengmingAirbnb The guideline says "Avoid `return` where not required." My interpretation of that guideline is "Don't use the `return` keyword in the last statement of a method," because Ruby doesn't require the keyword in the last line. It sounds like your interpretation was "Don't use the `return` keyword at all."

Perhaps the guideline could be rewritten as "Don't use the `return` keyword in the last statement of a method" if that was the original intent. I'd be fine with that, because I agree that sometimes an early `return` leads to more readable code with less deep indentation. Would you?

Regarding your examples, I think there's an underlying problem in the first two examples that, if fixed, would help to eliminate the need to discuss early returns: the examples contain "a long tedious block with 30 ~ 40 lines." In Ruby, that's typically considered too long for a method. (I hesitate to bring this up because I don't want to impose an upper limit in the style guide, but 30-40 is above any upper limit I've heard people discuss in Ruby.)

fengmingAirbnb:
@clizzin Got it. I think Gary's confusion as a reader mostly comes from the fact that the following code did not document well whether the author intends to return the reference or not.
```
def update(new_value)
    @value = func(new_value)
    Rails.logger.info '...'
end
```
In the OO world, returning an internal variable is a big deal since it breaks the encapsulation. I guess everybody understands that importance.
If the author does not intend to do so, he SHOULD put `nil` at the end. I think our code base are missing a lot of these `nil`s.

@BMorearty Thanks much for explanations! I totally agree with you that we should refactor these smelly codes. I just wanted to show that an early termination helps the readers a lot, for instance, when they try to do refactoring.

PR #45, opened 2015-04-01
Use bang, not not.

BMorearty:
cc: @romanfuchs @nelgau 

`not` falls in the same category as `and` and `or` and perhaps we should avoid its use.

If a good citizen is changing `and` and `or` to `&&` and `||` in existing code, but does not also change `not` to `!` in the same expression, they run a stronger risk of breaking existing logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ In either case:
     end
     ```
 
-* The `and` and `or` keywords are banned. It's just not worth it. Always use `&&` and `||` instead.
+* The `and`, `or`, and `not` keywords are banned. It's just not worth it. Always use `&&`, `||`, and `!` instead.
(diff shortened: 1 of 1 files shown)


romanfuchs:
:+1:

LogicWolfe:
:+1:

clizzin:
Looks good to me.

robotpistol:
:+1:

robotpistol:
I'm actually surprised this wasn't specified before.

wyattdanger:
:clap:

jasonkb:
LGTM

robotpistol:
i'm still suprised this wasn't specified before

PR #48, opened 2015-05-22
Updates to specification of Ternary use.

robotpistol:
* Add a suggestion for avoiding multiple conditionals in ternary statements
* Add examples for avoiding multiple conditionals in statements
* Add examples for avoiding mutli-line ternary statements

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,35 +643,12 @@ In either case:
     ```
 
 * Avoid multiple conditions in ternaries. Ternaries are best used with single conditions.
-  Convert these statements
-
-    ```ruby
-    # bad
-    some_condition && some_other_condition ? truthy_value : falsey_value
-
-    # good
-    if some_condition && some_other_condition
-      truthy_value
-    else
-      falsey_value
-    end
-    ```
 
 * Avoid multi-line `?:` (the ternary operator), use `if/then/else/end` instead.
     ```Ruby
     # bad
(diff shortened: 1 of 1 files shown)


BMorearty:
These seem like reasonable suggestions...but IMO, not worth including in the style guide. (I've been thinking it would be awesome if we try to limit rules to the most important ones, in the interest of keeping the guide from getting unwieldy.)

robotpistol:
You make a reasonable point.  My main intent was to add examples for an item already in the style guide about discouraging multi-line ternaries.  I don't mind dropping the section for multiple clauses (though I think they should also be discouraged)

BMorearty:
:+1: to dropping the section for multiple clauses. I don't think the examples are necessary for the other one but I don't object if folks think they are.

tommydangerous:
Close PR?

robotpistol:
I'm gonna fix this up slightly still feel strongly about multi line ternaries

robotpistol:
Ok, updated this PR.  I think that multi line ternaries and complex ternary conditionals negatively impact code.  They should be used sparingly.

I  do think that a number of the examples are redundant and unnecessarily didactic.
PTAL

ljharb:
LGTM if @BMorearty agrees

tommydangerous:
LGTM, I like.

BMorearty:
Ok :+1:

PR #51, opened 2015-10-12
Add a space after TODO

halilim:
IMO `TODO` and name run into each other without a space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ Thus when you create a TODO, it is almost always your name that is given.
   # TODO(drumm3rz4lyfe): Use proper namespacing for this constant.
 
   # good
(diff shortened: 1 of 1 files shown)


ljharb:
How does this interact with `rake notes`?

BMorearty:
Thanks for the pull request, @halilim! 

@ljharb, `rake notes` works fine with either format. 

Even though I would have written the guideline with a space if I had been the original author, I'm fine with it as-is and would like to leave it unchanged, considering the number of places in our code that use the prescribed style.

halilim:
OK, I guess one can always use a fork with the space :)

tommydangerous:
Close PR?

halilim:
Oh, sorry, I forgot about it.

PR #52, opened 2015-10-20
Prefer `some_method` over `self.some_method`.

BMorearty:
To: @clizzin 

Idiomatic Ruby does not use `self.` when making method calls. Without a guideline about it, people seem to get confused and over-use `self.`.

Purposely written as a "prefer" to accommodate the occasional case where `self.` really does add clarity.

Also updated code samples elsewhere in the doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,7 +356,7 @@ def fallbacks_for(the_locale, opts = {})
   #    children).
   # 1) The default locale is just a language. (Like :en, and not :"en-US".)
   if opts[:exclude_default] &&
-      ret.last == self.default_locale &&
+      ret.last == default_locale &&
       ret.last != language_from_locale(the_locale)
     ret.pop
   end
@@ -556,7 +556,7 @@ In either case:
     parts[i] = part.to_i(INTEGER_BASE) if !part.nil? && [0, 2, 3].include?(i)
 
     # okay
-    return if self.reconciled?
+    return if reconciled?
     ```
 
 * Never use `unless` with `else`. Rewrite these with the positive case first.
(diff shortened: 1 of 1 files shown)


ljharb:
👍

philihp:
:+1:

LogicWolfe:
:+1: /cc @robotpistol as I believe he frequently advises people to do the opposite of this and may wish to weigh in.

robotpistol:
I'm have no objections to this.  I think we use this often time to help make sense of methods that are longer than they should be.  It can also be nice on models to make it clear that your are assigning a value on the model itself.

BMorearty:
> I'm have no objections to this.

Great!

> I think we use this often time to help make sense of methods that are longer than they should be.

I agree that people probably do it for that reason. I think you would agree that a better remedy is to shorten the method. :-) 

> It can also be nice on models to make it clear that your are assigning a value on the model itself.

It's required for that case. I noted that in the PR as "left-hand side when calling an assignment method," but if you think it would be more clear, I can add the phrase, "including assigning an attribute when `self` is an ActiveRecord model."

robotpistol:
:+1:

BMorearty:
@robotpistol :+1: to the PR or :+1: to adding the wording?

justjake:
Why is idiomatic Ruby obscure?

robotpistol:
:+1:  to both

BMorearty:
Cool, change has been made. I'll wait for the gatekeeper @clizzin, to merge.

BMorearty:
Oops, @clizzin is out for a while. Merging.

jasonkb:
LGTM

PR #54, opened 2015-11-11
[private definition] Add a new line between private/protected and method definition

tommydangerous:
We should add a new line between `private` and `protected` and the method definition underneath it.

Example:
```
class Person

  def age
  end

  protected

  def dob
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -894,7 +894,7 @@ in inheritance.
     ```
 
 * Indent the `public`, `protected`, and `private` methods as much the
-  method definitions they apply to. Leave one blank line above them.
+  method definitions they apply to. Leave one blank line above and below them.
 
     ```Ruby
     class SomeClass
--- a/README.md
+++ b/README.md
@@ -903,6 +903,7 @@ in inheritance.
       end
 
       private
+
       def private_method
         # ...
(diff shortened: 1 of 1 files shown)


BMorearty:
If we're going to change the example, we'd better also change the wording above the example. The wording says leave one blank line _above_ but not below.

I'm in favor of this change (if the wording is changed too) but I'm fine with leaving it as is, too. I've forced myself to get used to it.

tommydangerous:
Updated the wording of the example to include below as per direction.

BMorearty:
I'm good with this change.

robotpistol:
LGTM

jasonkb:
Haha I don't understand why there is positive sentiment around this while there is negative sentiment around wanting "private" and friends to be indented one space less.

I personally supported both endeavours.

tommydangerous:
Why is this closed? I haven't made the change yet :(

BMorearty:
That was a mistake! Sorry.

BMorearty:
I blame jetlag

tommydangerous:
Glad you made it to Paris safely!

PR #56, opened 2015-11-11
[datetime] Prefer Date.current and Time.current

tommydangerous:
If we use `Time.now`, it will use the server's timezone.

If we use `Time.current`, it will look in the default timezone stated in the `config` file. If it is present, it will use that timezone. If it is not present, it will default to use the server's timezone.

With `Time.current` you get `Time.now` and some.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1133,6 +1133,10 @@ in inheritance.
     end
     ```
 
+* Prefer `Time.current` over `Time.now`
+
+* Prefer `Date.current` over `Date.today`
+
 ## Be Consistent
 
 > If you're editing code, take a few minutes to look at the code around you and


ljharb:
These should have more explanation in the text. Are these rails conventions?

robotpistol:
Also our current system is all UTC all the time.

robotpistol:
We convert to a local timezone when we intend to put it int the perspective of a specific user.

tommydangerous:
Updated comment.

I use in the code we use `Time.zone.now` and `Time.now` and some other variations. We can standardize the way we create `Time` or `Date` objects.

BMorearty:
I fear this is veering too much into "practices" rather than style. (There are a few other guidelines in this doc that do that, too.) If we start documenting practices, this doc could grow tenfold in size and be too hard to memorize all at once.

robotpistol:
I also feel like this is not really a style thing. Also it is very rails specific.

robotpistol:
Ah it is in the rails section at least.  I still think we need more justification as to why this is necessary as opposed to what we've been doing before.

tommydangerous:
I do agree with having too many guidelines, impossible to remember. I just see a lot of inconsistencies between us using `Time.now` and `Time.zone.now` and figured it would be nice to have 1 way we write our `Time` or `DateTime` objects. But at the cost of remembering too many and implement nothing, I agree we can leave this out.

robotpistol:
The inconsistencies I think are mostly because a section of our code has Rubocop that enforces using Time.zone.now

I don't know if the reasoning behind doing that.  All of our systems are UTC and expect UTC. In our current setup `Time.now` is functionally equivalent to `Time.zone.now`

LogicWolfe:
I think this falls into the category of not useful enough to warrant being in the guide. The longer the guide gets the less likely it will be people will read it, internalize it, and apply it consistently. :-1:

jasonkb:
@tommydangerous I agree with @LogicWolfe. I think the biggest opportunity here is for more Airbnb documentation and understanding of time zones and how it relates to Airbnb data models. This documentation is sorely lacking, or generally out of date, in our hackpad and code! Are you keen to add such documentation in Airbnb's documentation?

tommydangerous:
I like adding documentation, just tell me where :-D

PR #57, opened 2015-12-08
Update to latest from airbnb/ruby

justin808:


justin808:
This is reversed. Sorry about the noise.

PR #60, opened 2015-12-17
Don't use the return value of `=` in conditionals

robotpistol:
Even with parentheses setting variables in a conditional should be avoided.

It has a higher chance for errors and conflicts with the rule that we should not add unnecessary parens.  Also it is generally easy and more clear to do the assignment before the conditional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,9 +591,7 @@ In either case:
       end
     ```
 
-* Don't use parentheses around the condition of an `if/unless/while`,
-  unless the condition contains an assignment (see [Using the return
-  value of `=`](#syntax) below).
+* Don't use parentheses around the condition of an `if/unless/while`.
 
     ```Ruby
     # bad
@@ -606,10 +604,6 @@ In either case:
       ...
     end
 
-    # ok
-    if (x = next_value)
-      ...
-    end
     ```
 
 ### Ternary operator
@@ -707,11 +701,10 @@ In either case:
     end
     ```
 
-* Using the return value of `=` (an assignment) is ok, but surround the
-  assignment with parenthesis.
+* Don't use the return value of `=` in conditionals
 
     ```Ruby
-    # good - shows intended use of assignment
+    # bad - shows intended use of assignment
     if (v = array.grep(/foo/))
       ...
     end
@@ -721,10 +714,12 @@ In either case:
       ...
     end
 
-    # also good - shows intended use of assignment and has correct precedence
-    if (v = next_value) == "hello"
+    # good
+    v = array.grep(/foo/)
+    if v
       ...
     end
+
     ```
 
 * Use `||=` freely to initialize variables.


venantius:
+1

BMorearty:
-1. I like the shorthand and haven't found it to be troublesome.

README.md
```diff
@@ -721,10 +714,12 @@ In either case:
...
     end
 
-    # also good - shows intended use of assignment and has correct precedence
-    if (v = next_value) == "hello"
+    # good
+    v = array.grep(/foo/)
+    if v
       ...
     end
+
```
kmsun07: newline intentional?

README.md
```diff
@@ -606,10 +604,6 @@ In either case:
       ...
     end
```
kmsun07: newline intentional?

ljharb:
+1, things that are sometimes useful but often confusing are harmful.

BMorearty:
More context: my -1 is a mild opinion, not a strong one. I'll hold my strong opinions for other PRs. :smiley:

LogicWolfe:
+1 Local state is dangerous and I'd like to see its creation highlighted as clearly as possible.

kmsun07:
:+1:

venantius:
Merging.

BMorearty:
Here's an honest question for those who +1'd this change ( @robotpistol @venantius @ljharb @LogicWolfe @kmsun07 ). I'm writing some code that, without this style guideline, I would have written like this:

```ruby
  def segment
    if checkpoint_reservation?
      CHECKPOINT
    elsif (reservation = past_reservation(:interval => PAST_RESERVATION_INTERVAL))
      @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
    elsif (reservation = upcoming_reservation(:interval => UPCOMING_RESERVATION_INTERVAL))
      @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
    else
      NON_TRIP
    end
  end
```

This looks to me like a textbook example of where assignment in a conditional is the best solution. What is a good way to write this code without using assignment in conditionals, while still:

* keeping it concise and readable,
* shortcutting the logic so I don't call `past_reservation` or `upcoming_reservation` (both of which perform DB operations) unless necessary?

The only alternative I can think of is the following, which is less concise and, to my eye at least, definitely less readable:

```ruby
  def segment
    if checkpoint_reservation?
      CHECKPOINT
    else
      reservation = past_reservation(:interval => PAST_RESERVATION_INTERVAL)
      if reservation
        @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
      else
        reservation = upcoming_reservation(:interval => UPCOMING_RESERVATION_INTERVAL)
        if reservation
          @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
        else
          NON_TRIP
        end
      end
    end
  end
```

In this case, I don't see a good way to break the method up into smaller methods unless those smaller methods set instance variables as side-effects, which seems far worse to me than assigning in the `if` condition.

robotpistol:
How about
```ruby
return CHECKPOINT if checkpoint_reservation?

reservation = past_reservation(:interval => PAST_RESERVATION_INTERVAL)
return @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST if reservation

reservation = upcoming_reservation(:interval => UPCOMING_RESERVATION_INTERVAL)
return  @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST

NON_TRIP
```

BMorearty:
Thanks, @robotpistol. I'm glad I asked. I didn't think of that option.

BMorearty:
(Although I must say the `return ... ? : if ...` lines read a bit confusing to me.)

robotpistol:
Then go ahead and do this.
```ruby
reservation = past_reservation(:interval => PAST_RESERVATION_INTERVAL)
if reservation
  return @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
end

reservation = upcoming_reservation(:interval => UPCOMING_RESERVATION_INTERVAL)
if reservation
  return @user.id == reservation[:guest_id] ? TRIP_GUEST : TRIP_HOST
end
```

venantius:
Great question, @BMorearty, thanks for asking! Obviously we want to avoid code with ever-increasing nested else statements. @robotpistol, love the code examples :)

PR #63, opened 2016-01-27
Added example without hanging indentation.

akuhn:
Added an alternative example without hanging indentation since we seem to use both on our codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,19 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
       ...
     end
 
+    # good
+    def self.create_translation(
+      phrase_id,
+      phrase_key,
+      target_locale,
+      value,
+      user_id,
+      do_xss_check,
+      allow_verification
+    )
+      ...
+    end
+
     # bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,19 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
...
 
+    # good
+    def self.create_translation(
+      phrase_id,
+      phrase_key,
+      target_locale,
+      value,
+      user_id,
+      do_xss_check,
+      allow_verification
```
ljharb: wouldn't this need a trailing comma?

BMorearty: No, that's not allowed by Ruby here.

ljharb: ah oops, nvm then

venantius:
I approve of extra examples :)

akuhn:
Getting back to @ljharb comment about trailing comma. I think function calls allow trailing comma, might be good to distinguish this in the style guide. At the moment it just speaks about "function parameters" without distinguishing between function signature and function callsites. I'll make a PR.

PR #64, opened 2016-01-27
Use shorthand syntax for symbol keys.

akuhn:
Replaced all symbol keys with shorthand syntax using `s/:(\w+) =>/\1:/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,19 +173,19 @@ characters. Notice techniques like:
 ```ruby
 scope = Translation::Phrase.includes(:phrase_translations).
   joins(:phrase_screenshots).
-  where(:phrase_screenshots => {
-    :controller => controller_name,
-    :action => JAROMIR_JAGR_SALUTE,
+  where(phrase_screenshots: {
+    controller: controller_name,
+    action: JAROMIR_JAGR_SALUTE,
   })
 ```
 
 ```ruby
 translation = FactoryGirl.create(
   :phrase_translation,
-  :locale => :is,
-  :phrase => phrase,
-  :key => 'phone_number_not_revealed_time_zone',
-  :value => 'Símanúmerið þitt verður ekki birt. Það er aðeins hægt að hringja á '\
+  locale: :is,
+  phrase: phrase,
+  key: 'phone_number_not_revealed_time_zone',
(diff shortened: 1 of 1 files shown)


BMorearty:
@akuhn If you want this to be a guideline, I'd recommend adding an explicit guideline about it in addition to changing all the examples. (Personally I've come to prefer the old `=>` syntax but I'm ok with either.)

README.md
```diff
@@ -262,7 +262,7 @@ module Translation
   # Australian, New Zealand variations is provided.
   class PrimAndProper
     def initialize
-      @converters = { :en => { :"en-AU" => AmericanToAustralian.new,
+      @converters = { en: { :"en-AU" => AmericanToAustralian.new,
```
BMorearty: If this PR goes through, please also update the affected indentation (the next 3 lines). I'll only comment here, with the assumption you'll look for other cases where the indentation got messed up.

ljharb:
fwiw i am strongly not in favor of encouraging people to exclusively use shorthand syntax - for any non-symbol keys, you *must* use hash rockets, and inconsistency within the same hash is gross enough that the whole affected hash thus should use hash rockets - also, symbol keys + symbol values looks wonky in the shorthand syntax.

Primarily the fact that "shorthand syntax does not work in all cases but hash rockets do" is why I prefer only ever using hash rockets (but certainly not proscribing shorthand)

robotpistol:
Completely agree with @ljharb 
Until shorthand is at feature parity with hash rockets.  They can not be a true replacement and should not be mandated.

venantius:
Can't say I'm a fan of this change either.

tommydangerous:
Sadly, as much as I use short hand syntax 99.9999% of the time, they cannot do everything the rocket does. I still use rockets in some cases; when the key is a number or an object. As much as I like this change I don't think there is a strong enough argument for it besides "It looks better", "I like it", "Takes less keystrokes".

tommydangerous:
I like this PR but this sadly contradicts @BMorearty PR on using hash rocket :(

ljharb:
Closing for now, until such time as our consensus changes.

akuhn:
True enough, the arguments in favor of hashrocket seem convincing.

Thank's for closing this.

PR #65, opened 2016-01-29
Preferred collection methods

venantius:
For a bit of background context, refer to PR 36290 on monorail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -986,11 +986,13 @@ in inheritance.
 
 * Prefer `map` over `collect`.
 
-* Prefer `detect` over `find`.
+* Prefer `detect` over `find`. The use of `find` is ambiguous with regard
+  to ActiveRecord's `find` method - `detect` makes clear that you're
+  working with a Ruby collection, not an AR object.
 
 * Prefer `reduce` over `inject`.
 
-* Prefer `size` over either `length` or `count`.
+* Prefer `size` over either `length` or `count` for performance reasons.
 
 * Use symbols instead of strings as hash keys.
 
--- a/README.md
+++ b/README.md
@@ -984,6 +984,10 @@ in inheritance.
   is a hybrid of `Array`'s intuitive inter-operation facilities and
   `Hash`'s fast lookup.
 
+*  Prefer `map` over `collect`, `detect` over `find`, and `reduce` over
+   `inject` and size over length. The reason the use of `detect` is preferred
+   over `find` is because of the ambiguity with ActiveRecord's `find` method.
+
 * Use symbols instead of strings as hash keys.
 
     ```Ruby
--- a/README.md
+++ b/README.md
@@ -984,9 +984,13 @@ in inheritance.
   is a hybrid of `Array`'s intuitive inter-operation facilities and
   `Hash`'s fast lookup.
 
-*  Prefer `map` over `collect`, `detect` over `find`, and `reduce` over
-   `inject` and size over length. The reason the use of `detect` is preferred
-   over `find` is because of the ambiguity with ActiveRecord's `find` method.
+* Prefer `map` over `collect`.
+
+* Prefer `detect` over `find`.
+
+* Prefer `reduce` over `inject`.
+
+* Prefer `size` over either `length` or `count`.
 
 * Use symbols instead of strings as hash keys.
 


README.md
```diff
@@ -984,6 +984,10 @@ in inheritance.
   is a hybrid of `Array`'s intuitive inter-operation facilities and
   `Hash`'s fast lookup.
 
+*  Prefer `map` over `collect`, `detect` over `find`, and `reduce` over
```
tommydangerous: What do you think about this:
```
* Prefer `map` over `collect`
* Prefer `detect` over `find`
* Prefer `reduce` over `inject`
* Prefer `size` over `length` and/or `count`
```

tommydangerous: Perhaps we don't need the reasoning since it will add more reading material lol

If people question the reason, we can link them to a PR discussion.

venantius: Yeah, in the original discussion people seemed to feel it was self-evident that `map` was better than `collect`. We didn't talk about `size` v. `length` and/or `count` in that PR but I'm happy to add them as well.

tommydangerous: I'm thinking we can leave out the reasonings behind the style guide so that the guide is less wordy and easier to digest.

kmsun07: not suggesting we add it, but can someone give me the one-line version of the whys? is it just arbitrarily choosing one? specifically the first 3 (i know that size/length/count actually have behavioral implications)

venantius: detect over find is outlined in the actual pr. Maybe @tommydangerous can outline the preference for map, inject, etc.

BMorearty: I like @tommydangerous's bulleted version a bit more. Seems more digestible.

Overall, I think these are good suggestions.

Re: @kmsun07's question about whether the first 3 are arbitrary (and no, I don't have numbers to back these up; it's just based on what I've seen at different companies and open source projects):

* I'm pretty sure the Ruby community uses `map` far more often than `collect`, so it's in keeping with community conventions (and that's one of the goals of this style guide).
* I think the Ruby community uses `find` slightly more than `detect`, but I am in favor of preferring `detect`, as suggested in this PR, to make it clear it's not an ActiveRecord call. That's an actual practical reason as opposed to just personal preference.
* I think the Ruby community uses `inject` slightly more than `reduce`, but you know, there's map/reduce and all that, so it seems kind of funky to use `inject`. Besides, `reduce`/`inject` is not used nearly as often as `map`/`collect`.

venantius: @schleyfox seemed to feel that there was a reason why `map` was clearly superior to `collect` in the original PR, but I don't know what the rationale was. 

I think, per @BMorearty's comments, the `inject`/`reduce` distinction is the one that people have the softest preferences around. I like `reduce` a little more because it's a clear parallel to a programming idiom in other languages, whereas I'm not aware of `inject` having this meaning outside of Ruby.

ljharb: fwiw `map`/`reduce` are also the terms used in JS, so frontend and ruby devs will be able to share the same terminology (would be true with `find` too but i think the ActiveRecord hazard trumps that)

venantius:
I've updated the PR to be in line with @tommydangerous's suggestions; PTAL?

README.md
```diff
@@ -984,6 +984,14 @@ in inheritance.
   is a hybrid of `Array`'s intuitive inter-operation facilities and
   `Hash`'s fast lookup.
 
+* Prefer `map` over `collect`.
+
+* Prefer `detect` over `find`.
```
ljharb: i still think the ActiveRecord explanation is necessary here, especially since "find" is more common in Ruby (and also is used in JS).

The rest are idiomatic ruby, so I don't think they need justification.

venantius:
Updated to provide rationale.

ljharb:
LGTM!

venantius:
Boom!

README.md
akuhn: I am late to this conversation but would like to add a bit of historical context. Ruby's collections originate from Smalltalk (the first language with collections) and that is where the `collect, inject, select, detect` names are taken from. It now seems to be our rational to prefer to functional names `map, reduce` because they are more familiar with today's generation of programmers—except when these names may lend themselves to confusion with Rails as is the case with `find, find_all`.

PR #68, opened 2016-02-10
[tommy] Change Ruby to ruby in syntax highlighting

tommydangerous:
to:
cc:
priority: low
size: small

Contents:
- When highlighting blocks of code, instead of using `Ruby`, we should use `ruby` since that is typically what Github prefers (less keystrokes for you as well :-D)


robotpistol:
sure.  LGTM

venantius:
haha, okay :+1:

venantius:
Merging because harmless :)

PR #69, opened 2016-02-10
[tommy] Add rules for array literals

tommydangerous:
PR: https://git.musta.ch/airbnb/airbnb/pull/36234
priority: low
size: small

Contents:
- Only use `%W()` vs `%w()` when you need interpolation


BMorearty:
:-1: Why legislate this? I'd rather leave this out.

Also the rule is ambiguous on the question: is this bad? `["Bob", "John", "Ragna"]`

tommydangerous:
I agree with you @BMorearty that this rule is ambiguous because all it says is for us to NOT use `%Q` when interpolation isn't present. I am fine with leaving it out, I am just following the merged PR's.

PR #70, opened 2016-02-10
[tommy] String literals

tommydangerous:
PR: https://git.musta.ch/airbnb/airbnb/pull/36233
priority: low
size: small

Contents:
- Only use `%Q()` vs `%q()` when you need interpolation


BMorearty:
Please no! :-1: This deviates far from the idiomatic norm.

I see no reason for this. Just use `%Q` when you have quotes in the string. It's just normal practice.

BMorearty:
P.S. We already have this rule which I think is sufficient and has good examples: "Use %() for single-line strings which require both interpolation and embedded double-quotes. For multi-line strings, prefer heredocs."

PR #71, opened 2016-02-10
Use hashrocket instead of JSON syntax for hashes.

BMorearty:
Rationale: see discussion for PR #64.

cc @akuhn @ljharb @robotpistol @venantius @tommydangerous

### DIFF
--- a/README.md
+++ b/README.md
@@ -994,14 +994,14 @@ in inheritance.
 
 * Prefer `size` over either `length` or `count` for performance reasons.
 
-* Use hashrocket syntax instead of JSON syntax for hashes.
+* Use hashrocket syntax instead of JSON syntax for hashes. Hashrocket works no matter what type the key is, and symbol keys + symbol values look wonky in the JSON syntax.
 
-    ```Ruby
+    ```ruby
     # bad
-    hash = { one: 1, two: 2, three: 3 }
+    hash = { one: 1, two: 2, action: :update }
 
     # good
-    hash = { :one => 1, :two => 2, :three => 3 }
+    hash = { :one => 1, :two => 2, :action => :update }
     ```
 
 * Use symbols instead of strings as hash keys.
--- a/README.md
+++ b/README.md
@@ -994,6 +994,16 @@ in inheritance.
 
 * Prefer `size` over either `length` or `count` for performance reasons.
 
+* Use hashrocket syntax instead of JSON syntax for hashes.
+
+    ```Ruby
+    # bad
+    hash = { one: 1, two: 2, three: 3 }
+
+    # good
+    hash = { :one => 1, :two => 2, :three => 3 }
+    ```
+
 * Use symbols instead of strings as hash keys.
 
     ```Ruby


tommydangerous:
But I prefer to use the short syntax though, I just don't think we should make short syntax a rule. But I do prefer it over hash rocket...

tommydangerous:
How about a rule saying if your key is a symbol, you may use short syntax? I like that!

BMorearty:
A guideline like that is already implied in the current style guide.

tommydangerous:
I see this rule: `Use symbols instead of strings as hash keys.`

But with this PR's rule, we can never use short syntax. I would prefer if we can still use the short syntax.

BMorearty:
These are guidelines, not hard rules, and "Use symbols instead of strings as hash keys" is a good guideline but one that cannot practically be followed 100% of the time. (In particular, when interacting with other systems.)

Can we wait for others to weigh in?

ljharb:
I think this is great - could we add more of the rationale? Specifically, that hash rockets are required for non-symbol keys, consistency within the same hash/method/class/file is important, and this rule means people don't have to think about the cases where one works and the other doesn't?

BMorearty:
@ljharb I'd rather we not add rationale as a habit because it makes the whole doc much too long, and if people want rationales they can look at the PRs. But I think it's fair to add a short rationale to this one. I'll see if I can come up with something super-concise.

> consistency within the same hash/method/class/file is important, 

That one is addressed [here](https://github.com/airbnb/ruby#be-consistent).

tommydangerous:
I agree with not adding rationale to the rules because it makes the doc a lot longer and harder skim through.

This is one rule I might break lol :-D

BMorearty:
Rationale added. PTAL.

ljharb:
We should have a longer offline discussion about that - I think that a guide without rationales is far less useful, and "long" isn't a problem if you add a table of contents (this applies to the JS style guide, and nobody's complained about it being too long ;-) ) "Skimming" through the guide isn't a use case I think we should encourage - style is critically important, and time should be dedicated to reading it in its entirety.

fwiw, the one you linked to certainly covers "consistency" in spirit, but it feels like kind of a stretch that I'd be able to remember that paragraph and make the mental leap that it applies in every situation.

BMorearty:
P.S. I'm perfectly willing to remove the rationale if most agree that we should avoid rationales.

ljharb:
Thanks, I think your changes are a marked improvement. :+1:

tommydangerous:
@ljharb I really like Thoughtbot's style guide because it is straight to the point (https://github.com/thoughtbot/guides/tree/master/style). It is easy to memorize and read. Sometimes I do wonder about the rationale but then I go into their PRs and search for the reason which can take long. I do like knowing the reason for things but I also like short and concise style guides.

robotpistol:
I used to think that new style hashes were the way to go but their dealing with their limitations over time have made me come to appreciate hash rockets more and more.  I'm in favor of adding this to the guide.

venantius:
:+1: to this! Thanks Brian :)

tommydangerous:
This is the only rule I don't agree with in our style guide.

tommydangerous:
I am deeply sad about this. We shouldn't have made this a hard rule... :-(

BMorearty:
@tommydangerous :sad: For your sake I considered making it a "prefer" or "avoid" guideline, but I couldn't figure out a circumstance under which it would make sense to deviate from the default style. E.g., if we had chosen JSON syntax we could easily make it a "prefer" guideline because there are times when the hashrocket is the only style that works.

tommydangerous:
I am still torn on this. I do not like how this is a hard rule "Use...".

I mostly deal with hashes that use symbols. I rarely store integers or objects as keys, and when I do I am aware to use hash rocket. Can we make this a soft rule, perhaps saying "Prefer using hash rocket syntax, unless you know what you are doing". Something along those lines.

ljharb:
You may, but the next person who comes along and edits the hash to have a non-symbol key (or a symbol value, which looks gross in JSON syntax) may not. Besides individual preference and saving 2 chars per hash entry, what would be the benefit to using JSON syntax ever?

LogicWolfe:
I'm not at all a fan of rules of this nature. I fail to see how making a prescribed choice adds meaningful clarity to the code here, and I'm firmly a believer of not having unnecessary rules. The fewer things we specify in our style guide the easier it will be to internalize the rules and to generate buy in that the styles we have chosen have been picked for good rather than arbitrary reasons.

I thought the intent was to initiate a significant rediscussion of the ruby style guide with an initial survey, but I've seen this repo be incredibly active lately. I've been holding off on engaging with it because I don't want to spend a bunch of time debating something that is just going to be replaced by a new system. Does anyone have insight into how the efforts that are happening in this repo line up with those broader and more inclusive efforts?

venantius:
Re: the rule in question - I'm afraid I don't see much value in having the JSON syntax available. We already overwhelmingly favor hash rockets in the codebase, and as has been discussed previously the JSON syntax is not at feature parity with hash rockets, which means there's an opportunity for new engineers to be confused about when it's okay to use them. I have trouble seeing a justification for both.

Re: @LogicWolfe's concerns - I'm the one running the survey for the style guide changes. Everything that's been happening on the style guide lately is unrelated to the survey's efforts, and the things that have had overlap I've explicitly asked the PR / issue owner to close so that we could cover them in the survey. I know the survey is taking a while to come together but formatting questions on the survey so the rules are clearly worded and have demonstrable examples is taking time (and, like all of us, I have other deliverables I'm executing against as well.)

BMorearty:
> ... and I'm firmly a believer of not having unnecessary rules. The fewer things we specify in our style guide the easier it will be to internalize the rules and to generate buy in that the styles we have chosen have been picked for good rather than arbitrary reasons.

@LogicWolfe I agree with you. Strongly. I've been saying the same thing, but you said it well and concisely. I don't want a huge number of guidelines that are hard to memorize--or worse, a fast-growing list, which is even harder to memorize. 

Because these are style guidelines, not coding practices, I feel like many of them don't add meaningful clarity to the code. I'm trying to understand why this one does so less than others.

I'm not insistent on keeping this guideline. I just thought it would be good to propose it since it seems weird to keep using both syntaxes all over. So I wanted to see what others thought.

If there's enough agreement to revert this and include it in the survey, I'd have no problem with that. Or even revert it and *not* include it in the survey, and allow people to continue using both styles. Because as I said, I'm strongly in favor of having fewer guidelines, not more.

LogicWolfe:
@BMorearty Thanks, I'm glad we agree on the that principle. I don't know that this rule is necessarily worse than others that have been proposed, I haven't been actively keeping up with the discussions here. I just happened to see this one and that it had been merged with strong support and felt like an important cost had been moved past too quickly in the evaluation of it.

My hope in commenting was less to have an impact on this specific rule and more to encourage us to highly prioritize simplicity in our style guide.

With regards to this specific rule - while a developer may at some point only have interacted with one form or the other of hash literals, both syntaxes are immediately clear in what they are describing, some type of key value pairing. It may be that a developer on seeing the new syntax for the first time might need to look it up to see if it implies any special behavior, but in reading the code neither form is ambiguous or unclear. I don't think there is a significant onboarding cost or code clarity cost to removing this rule. I'm not suggesting we prefer json style syntax, just that we don't specify a preferred syntax.

@venantius One thing I would love to see in the survey is that any proposed rule has an option for the preference of we just make no rule at all, rather than only being able to choose between various forms of the rule.

venantius:
@LogicWolfe to be clear - the survey is not choose A or B, it's "add rule or do not"

LogicWolfe:
@venantius :+1: great!

venantius:
@LogicWolfe The concern with this rule isn't that the rule isn't readable, it's that a developer might observe the JSON syntax in use for symbol keys and assume it is acceptable to use JSON syntax with basically anything else, which is not supported. The JSON syntax support feels like a very half-baked feature to me, personally.

I don't have much else to contribute to this conversation beyond what's already been said.

akuhn:
As a recent convert to hashrockets I would like to bring forth searchability as another argument in favor. Consistent use of hashrockets allows symbols to be found with a simple `:name` query rather than the clunky `/:?name:?/` regular expression. Which comes in particular handy when codesearch is triggered with a keyboard shortcut that uses the current editor selection as query.

PR #78, opened 2016-02-12
Adding links to rules

### LINKED ISSUE #75: Link to rule
One of the awesome things about bbatsov's guide is that each rule is accompanied by a link to that particular rule, so that you can easily reference it in a discussion. We should do the same thing here!

venantius:
Resolves #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,11 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
 
 ### Indentation
 
-* Use soft-tabs with a two space-indent.
+* <a name="default-indentation"></a>Use soft-tabs with a two
+    space-indent.<sup>[[link](#default-indentation)]</sup>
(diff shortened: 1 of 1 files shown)


venantius:
(man, was this boring)

BMorearty:
Spot-tested by going to https://github.com/airbnb/ruby/tree/link-to-rule and clicking a few of them. It was so boring. :stuck_out_tongue: Seems to work, although some of the large sections at the beginning are missing a link (maybe because they are implicitly linked because they are headers).

Speaking of headers, turning every rule into a `####`-style header seems like an easier way to do it. But I haven't tried it. Maybe it wouldn't look good.

tommydangerous:
:+1:

venantius:
Per #75, I'm going to assume this isn't contentious and am going to merge now.

PR #79, opened 2016-02-19
[tommy] Moved indentation examples into rationales

tommydangerous:
to: (not sure who to tag on this)
cc:
priority: low
size: small

Contents:
- Extract the examples out of the `README` and into the `rationales`

Screenshots:
- `README`
screenshot 2016-02-18 17 57 40

- `rationales`
screenshot 2016-02-18 17 57 52

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,87 +41,21 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
 
 ### Indentation
 
-* <a name="default-indentation"></a>Use soft-tabs with a two
-    space-indent.<sup>[[link](#default-indentation)]</sup>
-
-* <a name="indent-when-as-case"></a>Indent `when` as deep as `case`.
-    <sup>[[link](#indent-when-as-case)]</sup>
-
-    ```ruby
-    case
-    when song.name == 'Misty'
-      puts 'Not again!'
-    when song.duration > 120
-      puts 'Too long!'
-    when Time.now.hour > 21
-      puts "It's too late"
-    else
-      song.play
-    end
-
-    kind = case year
-           when 1850..1889 then 'Blues'
-           when 1890..1909 then 'Ragtime'
-           when 1910..1929 then 'New Orleans Jazz'
-           when 1930..1939 then 'Swing'
-           when 1940..1950 then 'Bebop'
-           else 'Jazz'
-           end
-    ```
-
-* <a name="align-function-params"></a>Align function parameters either all on
-    the same line or one per line.<sup>[[link](#align-function-params)]</sup>
-
-    ```ruby
-    # good
-    def self.create_translation(phrase_id,
-                                phrase_key,
-                                target_locale,
-                                value,
-                                user_id,
-                                do_xss_check,
-                                allow_verification)
-      ...
-    end
-
-    # good
-    def self.create_translation(
-      phrase_id,
-      phrase_key,
-      target_locale,
-      value,
-      user_id,
-      do_xss_check,
-      allow_verification
-    )
-      ...
-    end
-
-    # bad
-    def self.create_translation(phrase_id, phrase_key, target_locale,
-                                value, user_id, do_xss_check, allow_verification)
-      ...
-    end
-    ```
-
-* <a name="indent-multi-line-bool"></a>Indent succeeding lines in multi-line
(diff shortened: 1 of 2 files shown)


venantius:
Hmmm. In general, I'd prefer to have examples still in the main README, since I find them really valuable. But I would endorse refactoring the examples you've specified to be shorter, per #80.

tommydangerous:
But look how sexy this is: https://github.com/thoughtbot/guides/blob/master/style/ruby/README.md

venantius:
While I like the brevity of Thoughtbot's guide, I actually find it hard to read visually. It also doesn't make clear to me what each rule *means*, which is why I like examples.

tommydangerous:
Well, I knew this would get lots of pushback. It was only a proposal. Maybe I'll try again next week :-)

venantius:
I mean, I'm not the final arbiter here. If other people feel strongly that we should do this then I'm happy to be overruled. But I do have an opinion :)

akuhn:
I second @venantius, a style guide without examples is hard to read. Language can be ambiguous, code example are helping to clear this up. In particular when given as combination of `#bad` and `#good` examples.

tommydangerous:
@BMorearty @ljharb @robotpistol @jasonkb @kmsun07

ljharb:
I also prefer examples in the main guide. In general I don't agree that a smaller style guide is better - imo the purpose is to be an authoritative reference, not something engineers have to, or will, memorize or internalize.

BMorearty:
@tommydangerous I'm against this. While I like the brevity of the Thoughbot style guide because it's easy to scan, I don't think it's appropriate to move a few code samples over to the rationale one at a time. Personally I still think I would prefer that this style guide keep the code samples inline with the guidelines.

Re: @ljharb's comment:

> the purpose is to be an authoritative reference, not something engineers have to, or will, memorize or internalize.

I strongly disagree. Engineers must be able to (but not be required to) memorize and internalize a style guide without too much trouble. Otherwise it's just bureaucracy that slows people down every day.

venantius:
I've been coming to the realization that the belief that the style guide should or shouldn't be something that is easily memorized is ultimately the root of many disagreements about what the best thing for the style guide is in the long term. 

I happen to agree with @ljharb, largely because I believe there are adequate tools to inform me about and/or correct style guide violations without my having to leave my development environment (and consequently, to have to pay much active attention to the style guide). Over time, I end up internalizing these rules, but I don't attempt to memorize all of them right away and instead am comfortable relying on the tool for guidance.

ljharb:
¯\\\_(ツ)_/¯ We're a team of over 300. Reading and maintaining code is far more important than writing it (and happens exponentially more often) - it's *OK* if you're slowed down while writing code - but it's not OK if style inconsistencies in the codebase slow down *reading* and *understanding* code.

BMorearty:
That's a false assumption, leading to a false conclusion. It's quite possible to to be fast in writing, reading, and maintaining code. And an enormous rulebook is the opposite of the way to do it.

tommydangerous:
This was an experiment. Thank you for the input and great discussion. Will close PR.

kmsun07:
i agree with jordan, reading greatly trumps writing. i'm also personally on board with the giant rulebook; no problem with not internalizing the whole thing upfront; we have automated tools, code reviews, and even past that, one slipping by isn't a big deal. just because people may not ever achieve 100% natural proficiency (which i don't necessarily believe) doesn't mean it's pointless to strive for standardization

akuhn:
Supporting that. I think it makes more sense to optimize a style guide for clarity (that is providing bad and good examples) rather than for brevity.

BMorearty:
>  I think it makes more sense to optimize a style guide for clarity (that is providing bad and good examples) rather than for brevity.

Just to clarify, I agree that the style guide should provide good and bad examples. I had the impression the conversation had veered off into a discussion of how many guidelines there should be, as opposed to how well-documented each guideline should be. But I may have misinterpreted the earlier statement.

akuhn:
Good point, these are different aspects. I guess aiming for a short list of well documented guidelines is best. Personally I feel many style guidelines are corollaries that follow from basic practices (like grepability, blameability, etc) and would thus favor a style guide that starts off by talking about those basic principles first.

PR #82, opened 2016-02-20
Shortening some examples to the minimum demonstrative length

venantius:
I've tried to tighten up a number of our examples to the shortest length that
reasonably captured the rule they were demonstrating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,19 +51,11 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
     case
     when song.name == 'Misty'
       puts 'Not again!'
-    when song.duration > 120
-      puts 'Too long!'
-    when Time.now.hour > 21
-      puts "It's too late"
     else
       song.play
     end
 
     kind = case year
-           when 1850..1889 then 'Blues'
-           when 1890..1909 then 'Ragtime'
-           when 1910..1929 then 'New Orleans Jazz'
-           when 1930..1939 then 'Swing'
            when 1940..1950 then 'Bebop'
            else 'Jazz'
            end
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,10 +68,6 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
     # good
     def self.create_translation(phrase_id,
                                 phrase_key,
-                                target_locale,
-                                value,
-                                user_id,
-                                do_xss_check,
```
BMorearty: This one looks funky because after removing these params, they would all fit on one line.

README.md
```diff
@@ -88,18 +76,14 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
...
-      user_id,
-      do_xss_check,
       allow_verification
     )
       ...
     end
 
     # bad
-    def self.create_translation(phrase_id, phrase_key, target_locale,
-                                value, user_id, do_xss_check, allow_verification)
```
BMorearty: same here

README.md
```diff
@@ -88,18 +76,14 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
     def self.create_translation(
       phrase_id,
       phrase_key,
-      target_locale,
-      value,
-      user_id,
-      do_xss_check,
```
BMorearty: same here

BMorearty:
LGTM except for the funky ones.

venantius:
@BMorearty do you think there's a length that's shorter than the original for the funky example that is still demonstrative? I know it could all fit on one line at the moment but I tried to shorten it to the minimal possible example that demonstrates the underlying principle. Just wondering how much I should really be re-adding to get the point across without taking up a screen full of space.

venantius:
We could also probably trim a few lines in that example by not forcing the rule to only apply to method definitions, since the same rules presumably apply to method calls as well.

BMorearty:
If joining all the lines would be >100 chars, then you can split the lines and show the result in the sample. At least that's what I would do.

tommydangerous:
Looks nice

venantius:
I'm going to close this since it's not super important.

PR #83, opened 2016-02-23
Revert "Merge pull request #71 from airbnb/brian-morearty-hashrocket"

BMorearty:
I regret ever suggesting this guideline (or any guideline about which hash syntax we prefer) a week ago. It was foolish consistency--consistency for its own sake. It doesn't help anyone. I'd like to revert it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -978,16 +978,6 @@ In either case:
 * <a name="size-over-count"></a>Prefer `size` over either `length` or `count`
     for performance reasons.<sup>[[link](#size-over-count)]</sup>
 
-* Use hashrocket syntax instead of JSON syntax for hashes. Hashrocket works no matter what type the key is, and symbol keys + symbol values look wonky in the JSON syntax.
-
-    ```ruby
-    # bad
-    hash = { one: 1, two: 2, action: :update }
-
-    # good
-    hash = { :one => 1, :two => 2, :action => :update }
-    ```
-
 * Use symbols instead of strings as hash keys.
 
     ```ruby


ljharb:
Why? I think it's very useful - and consistency is valuable, even for its own sake.

BMorearty:
Because I don't see how it's useful, and I don't think consistency is valuable for its own sake. See earlier points brought up by @LogicWolfe [here](https://github.com/airbnb/ruby/pull/71#issuecomment-185293802).

ljharb:
That's one comment amidst a thread of people who were mostly in favor of it. Many of us *do* see how it's useful, and do think consistency is valuable for its own sake, as has been demonstrated.

BMorearty:
> as has been demonstrated

Where was it demonstrated that consistency is valuable for its own sake?

I don't want to introduce guidelines that I think are superfluous.

ljharb:
Sorry, to clarify, I think it's been demonstrated that many of us believe that.

tommydangerous:
I would prefer this not be a hard rule. I rarely use objects or integers as keys for my hashes.

ljharb:
or strings? we use that very often in our codebase.

tommydangerous:
Forgot about strings, that to. I would like to be given the option to choose when I absolutely need to use hash rocket syntax since I cannot remember the last time I needed it while coding in Monorail. I understand we use it all over Monorail, but personally I haven't had to store anything but symbols as my keys lately.

venantius:
As noted in the original email, I worry that the JSON syntax's lack of proper support for all object types as keys will lead to confusion from unfamiliar engineers trying to mimic it when they come across it. For that reason, I do think preferring hashrockets is the right thing to do.

venantius:
Oh, and there was also @akuhn's point about hashrocket exclusivity being superior from a code search perspective.

BMorearty:
Re: confusion, anyone who works in a Ruby code base needs to learn this extremely basic aspect of Ruby syntax.

venantius:
I think there's enough contention around this to warrant the revert. While I'd be interested in revisiting this in the future in the context of the same sort of survey being conducted currently, I don't think the original PR's balance of 5 for, 3 against suggests that it belongs in the style guide currently. 

Feel free to continue to discuss, but you've got my :+1: on merging the reversion.

kmsun07:
seems odd to have this rule when we don't have a similar guideline about, say, single vs double quotes (which actually may a performance impact).

i personally prefer to use the json style just because ti's more concise, but i ALSO prefer standardization whenever possible - consistency for consistency's sake is helpful for readability. anyway, fine with reverting this; i'm not sure reading both styles of hashes actually introduces any confusion; if someone is confused by the difference, then they will likely be just as confused when they can't naturally use indifferent access for something defined by hash rockets

ljharb:
While I still prefer to leave this as a strict guideline, if we do decide to change it, could we at the least leave behind a soft guideline? Kind of a "prefer" versus the current "use"

kmsun07:
all style rules are soft guidelines. if we want to allow both, then we should straight up allow both

robotpistol:
A more reasonable rule of thumb is don't mix styles in a given call/hash.

venantius:
@kmsun07 not all guidelines are soft guidelines. We're planning on enforcing the style guide with RuboCop, so specific guidance on whether the guideline is soft or not does matter.

venantius:
I do agree with @robotpistol on not mixing styles in a given hash, though. It sounded like there wasn't any opposition to that in the other forums we've talked about this.

tommydangerous:
Single and double quotes only have noticeable performance differences when interpolation is involved.

tommydangerous:
This guide can say "Do not mix rocket syntax and short syntax in the same hash"
```
# bad
{ :light => "saber", dark: "side" }

# good
{ :light => "saber", :dark => "side" }

{ light: "saber", dark: "side" }
```

akuhn:
:+1: for not mixing (and even for a "prefer" guideline but no strong feelings about this part).

:+1: for single quoted strings unless interpolation is used.

kmsun07:
i don't particularly like that thinking, but if so, then all guidelines are HARD guidelines, and there's no place for a soft guideline in there

kmsun07:
@tommydangerous i thought that the fact that the interpreter needs to LOOK for interpolation itself is the cost - is ruby smart enough to not need to do that?

venantius:
I don't understand why that would make all guidelines hard guidelines. If we say "keep your variable names simple and clear" that's not something we would expect RuboCop to enforce, but it's a reasonable style guideline.

kmsun07:
rules like that don't belong in a ruby style guide, as they are not related to ruby

ljharb:
Sure they do - an Airbnb ruby style guide should include all things to consider when writing Ruby code at Airbnb, which includes more things than just ruby-specific things.

andykram:
:+1: for reverting

kmsun07:
hmm, i disagree; we should try not to get in the business of describing generally good programming principles; i think we should assume the reader is a competent programmer looking to understand ruby best practices, rather than someone learning to code from scratch

BMorearty:
Some of these are indeed soft guidelines, not hard rules. (This is a style _guide_, after all.) They can be identified with the words "prefer" or "avoid." 

Having said that, some engineers are tired of humans endlessly commenting on their style and would like just to have a bot tell them. So I'd be ok with using Rubocop and turning _some_ of these into style rules. I still think there are some guidelines in here that don't make sense as rules, though. One example: [File/class-level comments](https://github.com/airbnb/ruby#fileclass-level-comments). For example, a model that's a subclass of `ActiveRecord::Base`. The name of the class, and the fact that its superclass is `ActiveRecord::Base`, probably give you the context you need to get started. When they don't, go ahead and add a comment. [The Antidote to Bureaucracy is Good Judgment](http://nerds.airbnb.com/the-antidote-to-bureaucracy-is-good-judgement/).

Re: mixed styles in a single hash, I'm pretty sure everyone agrees that's not good style.

> Single and double quotes only have noticeable performance differences when interpolation is involved.

That is correct. And just like it's not valuable to specify which type of hash to use, it's also not valuable to specify which type of string delimiter to use. (Other than [the existing guideline](https://github.com/airbnb/ruby#percent-parens)). I've been in the Ruby community long enough to have heard this debate and seen the data. There's no difference. And even if there _were_ a small difference, micro-optimizations are not even worth discussing in the context of a web app, whose bottleneck is network calls to a database. 

> we should try not to get in the business of describing generally good programming principles

I agree with @kmsun07. This guide is about style, not practices or principles. This is something I've been trying to communicate: just think of how many good programming practices you know, that are language-and-framework agnostic. 1,000? 2,000? Putting them all in a Ruby style guide drowns out the value of the actual Ruby guidelines. It's as waste of everyone's time. Even Ruby-specific practices like the [Scopes section](https://github.com/airbnb/ruby#scopes) don't really belong here, because they are practices, not style. 

Everything in moderation, though. I'm sure someone could point out a non-Ruby-specific practice that I have supported having in the guide. :wink:

ljharb:
It's really unfortunate that this was merged without actually achieving consensus.

BMorearty:
@ljharb I'm sorry about that. I would definitely not have done that under most circumstances. This was different. It was (a) a revert (b) of something I had submitted (c) that was only about a week old. If any one of those three were not true, I would have waited for more discussion.

venantius:
I think the revert here was warranted. Although we seemed to have consensus at merge time on the original PR, it subsequently became apparent that we did not. Brian also changed his mind about its worthiness, which he's obviously allowed to do. I don't think the fact that we did something in the past should be a reason for us to stick with it (true in this specific case, and true in general for the style guide), and it's okay for us to recognize when we've made a mistake and to go back on it.

ljharb:
The initial request already achieved consensus - that you were the submitter and changed your mind is irrelevant - new consensus is required.

LogicWolfe:
Both myself and @tommydangerous expressed reluctance in the original PR. The merge happened before it became clear that there were objections, but I don't know it's fair to call that consensus.

ljharb:
Perhaps that's the case.

It seems that before we make any further decisions on or changes to specific rules, we need to agree on what a style guide *is* and is meant to be - and we should discuss that internally, rather than here on our public Github.

BMorearty:
I agree, it seems time to discuss that, and it should be an internal discussion.

PR #88, opened 2016-03-15
[survey] Add rules agreed on in survey

plberg:
Most content is identical to that in the survey. The text and examples for a few rules were updated based on comments in the survey and followup discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,24 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
     end
     ```
 
+* <a name="method-def-empty-lines"></a>Use a single empty line to break between
+    statements to break up methods into logical paragraphs internally.
+    <sup>[[link](#method-def-empty-lines)]</sup>
+
+    ```ruby
+    def transformorize_car
+      car = manufacture(options)
+      t = transformer(robot, disguise)
+
+      car.after_market_mod!
+      t.transform(car)
+      car.assign_cool_name!
+
+      fleet.add(car)
+      car
+    end
+    ```
+
 ## Line Length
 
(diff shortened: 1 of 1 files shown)


ljharb:
it would be ideal for this repo to have a rubocop config, and for each applicable section to link to the rubocop documentation for the applicable rules (that doesn't have to be in this PR tho)

romanfuchs:
I think all of these rules make sense and it's great to have agreed on them to be included in Rubocop, but I think a lot of them don't need to be in the styleguide. Some of them seem too much common sense to be explicitly listed here. I'd rather avoid bloating this style guide too much.

Other opinions?

ljharb:
Personally I think that the styleguide should match 1:1 with our rubocop config, have as many rules in it as possible, and that "bloat" isn't really a concern. However, I think that this might be a discussion we should have internally to agree on what a styleguide should actually be.

tommydangerous:
I plus 1 the rubocop config. I agree with @romanfuchs on not bloating the style guide. I'd prefer a short and concise style guide vs an explanation for every single rule.

We can perhaps have 2 documents, 1 explaining all the rules and the actual style guide.

tommydangerous:
@plberg LGTM!

Per our discussion, I now believe getting all of these rules documented in our style guide is good.

Then as a future follow up, if we wanted to have a more concise, brochure-esque, style guide, we can have a separate document linking to our "Bible-esque" style guide with all the comprehensive rules and examples.

romanfuchs:
+1 to what Tommy said.

venantius:
LGTM

plberg:
Awesome! I love where this is headed.

PR #89, opened 2016-03-21
[fix] Fixing bad spacing on example

venantius:
The newline example should have two spaces for the multi-line conditional, not
one.

cc @plberg

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
 
     ```ruby
     if @reservation_alteration.checkin == @reservation.start_date &&
-       @reservation_alteration.checkout == (@reservation.start_date + @reservation.nights)
+        @reservation_alteration.checkout == (@reservation.start_date + @reservation.nights)
 
       redirect_to_alteration @reservation_alteration
     end


kmsun07:
i looked at the rest of the guide and there's no actual rule for this. shouldn't we add the rule specifying the actual rule before we "fix" it?

other spacing examples: https://github.com/airbnb/ruby#only-simple-if-unless and https://github.com/airbnb/ruby#block-and-inline-comments (the one you seem to be following)

closest thing to an existing rule: https://github.com/airbnb/ruby#indent-multi-line-bool

BMorearty:
I think @kmsun07 makes a good point.

His examples also point out that it might be worth having different guidelines for these two cases:

1. A multi-line conditional that starts with a keyword such as `if` or `while`. For these cases, we could discuss whether we like to align the subsequent lines with the first non-whitespace after the keyword or not.
2. A multi-line conditional that stands on its own, as a method return value. For these, the https://github.com/airbnb/ruby#indent-multi-line-bool guideline that @kmsun07 points out is a good guideline to keep.

venantius:
I think those are legitimate points and I'm definitely open to clarifying that point further. In this specific case, though, this seems to be a clear instance of a multi-line boolean within the context of a conditional keyword. Am I missing something?

BMorearty:
Isn't that the case that's worth discussing?

The case that's already clear is a multi-line conditional with no keyword. (Although that guideline could be clarified that it's talking about the case when there's no keyword. The example shows it.)

venantius:
Ah, I didn't see that our multi-line boolean example didn't have a keyword. That's what I was missing. 

I thought from past discussions about this that we wanted extra indentation when there was a keyword, since that was the case where it was hard to tell where the conditional ended and the block began (though I think the same argument could be made whether or not the conditional keyword was there). 

Are there differing opinions here?

BMorearty:
That sounds nice, if by "extra indentation" you mean "align with the first character after the whitespace."

```ruby
if foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo &&
   baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
````

and

```ruby
while foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo &&
      baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
```

Or did you mean:

```ruby
if foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo &&
    baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
````

and

```ruby
while foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo &&
    baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
```

kmsun07:
either of those seem fine to me. there's also the "complication" of what happens if the first condition itself spans multiple lines and/or requires nesting (mixing && and ||)

i personally find I tend to go out of my way to avoid having this situation arise, usually putting it in a helper function or calculating into a variable

tommydangerous:
Wait a second, I thought we align with the conditional on the top line...

ljharb:
fwiw, I prefer never aligning anything, and only using complete indentation levels to indicate semantic nesting - ie:
```ruby
if some_long_conditional &&
  some_other_long_conditional

  ...
end
```

Visual alignment often makes diffs more noisy as things are tweaked, as well as creates a subjective standard people would otherwise not have to debate about in code reviews.

venantius:
@BMorearty and @kmsun07 I meant the latter - an extra two spaces, beyond the normal indent for the conditional block. That would make the indentation follow the same indentation we use in the examples for https://github.com/airbnb/ruby#block-and-inline-comments and https://github.com/airbnb/ruby#indent-multi-line-bool

kmsun07:
i'm fine with that. should we codify? is there a process to get consensus?

rahul342:
FWIW2, I agree with @ljharb suggestion. It keeps thing simple and that coupled with https://github.com/airbnb/ruby#multiline-if-newline should offer readability.

BMorearty:
My preference is align with the first non-whitespace character after the keyword but I'm ok with any of the others. Maybe this should go in a survey later on?

tommydangerous:
My preference is align with first non whitespace character (same as @bmorearty)

plberg:
From looking at the style guide and examples in our codebase, there are actually 4 current styles.

2-space indent:
```
if long_foo? &&
  long_bar?
  …
```

4-space indent:
```
if long_foo? &&
    long_bar?
  …
```

Align with boolean:
```
if long_foo? &&
   long_bar?
  …
```

Align with boolean + 2-space indent:
```
if long_foo? &&
     long_bar?
  …
```

This same topic (multi-line conditional alignment) was addressed in a short conversation in PR #39288 in Monorail. Here is a quote of the verdict we reached there:

> We indent booleans in conditionals rather than aligning.

“Indent” here is referring to a standard, 2-space indent. This is the guideline I followed when cleaning up any multi-line conditionals I touched during code refactoring in Monorail.

I could be convinced to follow any of these styles. Also, while I agree with @kmsun07 that clunky multi-line conditionals warrant a larger refactor, they exist in the wild, and it is probably worth it for us to choose one style! However, I’m wary of settling on a single style here, in the context of this PR. The survey process worked really well, had positive feedback, and helped ensure all interested parties had an equal voice - perhaps we should file a ticket for this and include it in a second survey, whenever that should occur?

akuhn:
+1 for indentation rather than alignment.

Personally I prefer a style not used in our codebase. 
```
if (
  long_running_predicate &&
  long_running_predicate
)
  ...
end
```
Why? The closing bracket acts as a clear separator between predicate and function body and it maintains the basic principle that blocks of code are enclosed by equally indented lines of less indention.

venantius:
Man, I had no idea I was stumbling onto such uncharted territory when I opened this PR! I think @plberg's proposal here is probably the right one; although this doesn't seem to be an issue that people have strong feelings about it seems there are enough possible options for it that we might want to defer to the next round to surveys to figure out what we want to do.

BMorearty:
Yes, let's survey.

venantius:
Will close for now, @plberg can add it to the list for next survey.

jgonera:
+1 to @ljharb and @akuhn's suggestions.

PR #90, opened 2016-03-21
[cleanup] Remove unused block args (redundant)

venantius:
Our unused block args rule is an explicit subclass of a general rule that
already tells us to use `_` for unused variables. No need to repeat ourselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -974,17 +974,6 @@ In either case:
     one-liner scripts is discouraged. Prefer long form versions such as
     `$PROGRAM_NAME`.<sup>[[link](#no-cryptic-perl)]</sup>
 
-* <a name="unused-block-args"></a>Use `_` for unused block arguments.
-    <sup>[[link](#unused-block-args)]</sup>
-
-    ```ruby
-    # bad
-    result = hash.map { |k, v| v + 1 }
-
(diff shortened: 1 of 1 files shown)


BMorearty:
:+1:

tommydangerous:
Sorry I am blind, can you point me to the "general rule that tells us too use `_` for unused variables" please. Thank you.

venantius:
https://github.com/airbnb/ruby#throwaway-variables

tommydangerous:
Awesome, thank you.

Perhaps we can add 1 or 2 more examples to this:
```ruby
result = hash.map { |_, v| v + 1 }

# The following scenario typically arises when overriding methods
def prepare_create(data, _, _)
   log_data(data)
end
```

venantius:
@tommydangerous I thought you were a fan of minimizing the example volume? I feel like the current example is demonstrative enough for the moment.

plberg:
LGTM!

robotpistol:
LGTM

venantius:
Going to merge for now. If we really want extra examples we can revisit.

rahul342:
Thanks @venantius!

PR #91, opened 2016-03-22
[readme] Remove set-unique rule

rahul342:
Remove `set-unique` rule because it's not a matter of style and more of Computer science concept of using appropriate data structure for the relevant operations.

@venantius

### DIFF
--- a/README.md
+++ b/README.md
@@ -1216,11 +1216,6 @@ In either case:
 
 ## Collections
 
-* <a name="set-unique"></a>Use `Set` instead of `Array` when dealing with unique
-    elements. `Set` implements a collection of unordered values with no
-    duplicates. This is a hybrid of `Array`'s intuitive inter-operation
-    facilities and `Hash`'s fast lookup.<sup>[[link](#set-unique)]</sup>
-
 * <a name="map-over-collect"></a>Prefer `map` over
     `collect`.<sup>[[link](#map-over-collect)]</sup>
 


ljharb:
Why would that category of thing not be appropriate for a style guide? Newcomers to Ruby may not be aware of the underlying implementation differences.

rahul342:
@ljharb Set is an abstract data type and not specific to ruby. The implementation of Set as a Hash set too is the most common and present in all popular languages. Python's `set` is implemented as a hash set as well. Isn't it true of any language that if you have group of unique elements that you'd be doing a lot of lookups, then you should use a set data structure?

In fact this rule is vague when it doesn't talk about what "dealing with unique elements" actually entails. 

https://github.com/airbnb/ruby#string-concatenation feels like more a language nuance.

I don't feel strongly about this one. I saw it more as a cleanup.

kmsun07:
i'm on board with removing it, for the stated reason that a 'set' is not a ruby concept. i do get that newbies may not know the nuances of the language, but we shouldn't get into the business of describing everything you can accomplish with ruby here :)

ljharb:
¯\_(ツ)_/¯ i've seen many instances in ruby codebases of using Array when Set will do, particularly because Set (until recently) didn't exist as a concept in JS (and we still can't use Sets on airbnb.com because of browser compatibility) - so this one in fact might present an issue for frontend devs.

BMorearty:
:+1: for removing this, as it's not a matter of style. We could use the style guide to teach every data structure and CS concept, but I don't think that's what a style guide is for.

tommydangerous:
:+1: for removing

venantius:
While I understand @ljharb's concerns, I think that's an area where JavaScript is the exception rather than the rule. In general I agree with Rahul that sets should
be considered a core data structure and consequently a computer science // software engineering pattern rather than a matter of style.

akuhn:
On a related note, we might want to put `Array#|` and `Array#&` on the blacklist of our guide.

Those set operations are defined on `Array` class but they don't accept `Set` arguments. This just caused a bug in prod for me :expressionless:

venantius:
I'm going to merge this. I think there's decent consensus here that it isn't a style issue.

BMorearty:
@akuhn The `Array#|` and `Array#&` issue is not style-related either.

akuhn:
Under which heading would you file them then? It seem to make sense to share these learnings about core libraries so we don't repeat mistakes.

ljharb:
Style guides tend to be about much more than just style - they're more "coding guides" which include style, than solely being concerned with "style" (which is a subjective category anyways)

BMorearty:
> Under which heading would you file them then?

A list of language-specific document of "good practices," as opposed to style guidelines, is where I would suggest a coding practice should go.

> Style guides tend to be about much more than just style ...

Not the ones I've worked with. It's useful to separate the concepts of style and practice. Style guidelines can then be learned quickly without having to become a master of the language. Practices, by necessity, are learned gradually. It takes a lifetime to learn all the possible good coding practices.

rahul342:
+1 to Brian's comment. Although it's important to have this info somewhere like "Best Practices guide", but it's not in code style guide and can potentially cause information overload.

PR #92, opened 2016-03-24
[ERF] Distinguish def and calls because trailing comma.

akuhn:
Getting back to @ljharb comment about trailing comma in https://github.com/airbnb/ruby/pull/63

Function calls allow trailing comma but definitions don't, might be good to distinguish this in the style guide. At the moment it just speaks about "function parameters" without distinguishing between function signature and function callsites.

Reviewers

- @bmorearty 
- @ljharb 
- @venantius
- and who not …


ljharb:
LGTM

BMorearty:
I worry about allowing this because function calls don't require parentheses. Incidents happen when people put a comma at the end of an argument list and there's no parenthesis and the next line is a statement.

akuhn:
Interesting point. How common are multiline calls without parens in monorail? Is that even something we would want to encourage? In data tools we use parens for all multiline calls.

BMorearty:
I think they're common. The style guide says avoid parentheses if you don't care about the return value of a call. It doesn't mention whether it's multi-line, and I think this it the right guideline for idiomatic Ruby. For example, a `before_filter` call that lists several `:only` actions in an array, so it puts each on its own line.

akuhn:
Gotcha. I'll add a warning that thus applies to function calls with parens only.

I also see that we recommend parens for multiline calls even if no return value, even though the use of the word "especially" seem confusing

> If the method doesn't return a value (or we don't care about the return), parentheses are optional. (Especially if the arguments overflow to multiple lines, parentheses may add readability.)

akuhn:
Added a warning with what @bmorearty said.

BMorearty:
I'm still not a fan of this but am totally willing to hear what others think.

Survey?

akuhn:
It seem pretty safe for calls with parens. Would you want to me make the warning more explicit? 

Trailing commas are required to keep diff footprint small and to not leak blame to unrelated changes. Not requiring a trailing comma will blame me for an unrelated line when adding or removing parameters in method calls. This might not be a big issue for product teams where the chance is high that you are already the author of this line, but it matters to infrastructure teams who often do sweeping code changes. 

Motivating example see monorail PR #41571, if I would have had to remove trailing commas in this diff I would be blamed to 146 experiment treatments and would have required more manual cleanup after my automated script that removed matching lines.

tommydangerous:
I am not in favor of allowing a trailing comma in method arguments. A ruby method should take a few finite arguments. We use trailing commas in arrays and hashes because occasionally we add to that array or hash. However, I do not think it is good practice to endlessly add more arguments to a method definition.

akuhn:
Ruby arguments include `options` hash, those tend to change just as frequently as hash and array literals in my experience.

swac:
Was this ever resolved? I'm personally a fan of this style, and it looks like we're going to be encouraging a similar style in JavaScript soon, so I imagine there will be confusion about the inconsistency there.

BMorearty:
This was not resolved. Next time we do a style survey, let's include it.

Don't worry about inconsistencies between language style guides. They're different languages and they each have their own style.

robotpistol:
An argument against trailing commas for methods is that they should only be used in conjunction with parentheses.  I don't feel strongly about it though as long as they are only enforced with parens.

robotpistol:
Either way we decide rubocop has us covered.

akuhn:
@robotpistol confused, so what is the final take on this one? Is it good to merge?

BMorearty:
@akuhn I think when there's another survey let's include this.

venantius:
Marking for next year's Q1 survey.

PR #98, opened 2016-06-09
[ruby] Add example against { } in multi-line block

conzjiang:
The syntax rule that advises against using braces for multi-line blocks is stated, but there isn't an illustrated example listed, so it's easily overlooked by some who are skimming the guide for best practices. Added an example so that it's more obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -834,6 +834,18 @@ In either case:
 
     # good
     names.select { |name| name.start_with?("S") }.map { |name| name.upcase }
+    
+    # bad
+    names.each { |name|
+      puts name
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -834,6 +834,18 @@ In either case:
 
     # good
     names.select { |name| name.start_with?("S") }.map { |name| name.upcase }
+    
+    # bad
```
kmsun07: it seems like the example immediately before and immediately after these two are a pair (they have the same behavior); should we make sure to keep them adjacent?

PR #100, opened 2016-06-13
Add example against {} in multiline block

conzjiang:
couldn't figure out how to modify this patch https://github.com/airbnb/ruby/pull/98 for some reason :/

this pr keeps the correct examples grouped together, while adding a new example (see #98 for more details)

@kmsun07

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,6 +833,18 @@ In either case:
     names.each do |name| puts name end
 
     # good
+    names.each do |name|
+      puts name
+      puts 'yay!'
+    end
+
+    # bad
+    names.each { |name|
+      puts name
+      puts 'yay!'
+    }
+
+    # good
     names.select { |name| name.start_with?("S") }.map { |name| name.upcase }
 
     # bad


kmsun07:
LGTM. not related to the code we care about, but it's a little weird to do `puts` and `print` interleaved, right?

conzjiang:
@kmsun07 I use `print` when I'm printing out a newline because it looks strange to me to have a blank `puts` on one line.

```ruby
puts name
puts
puts "..."
```

If it's not in line with our style guide though, I can change it!

BMorearty:
LGTM too, but I agree with @kmsun07. Doing `print "\n"` seems really strange to me, even more so when mixed with a call to `puts`.

I don't think it's specified in the style guide. It's just what I think most people do.

kmsun07:
this isn't really relevant to the style guide; print/puts is rarely present in a production web app

that said, maybe if we just changed it to
```
puts 'line 1'
puts 'line 2'
```
it would be fine? the goal is just to not detract from the main purpose of the example, so the less 'interesting' the contents, the better

conzjiang:
@kmsun07 Makes sense! 😄  just made the change, i'll merge if the new change looks good to you

conzjiang:
@kmsun07 @BMorearty Makes sense! 😄  just made the change, i'll merge if the new change looks good to yall

kmsun07:
👍

BMorearty:
👍

conzjiang:
merged #100

PR #103, opened 2016-08-17
chinese version

1c7:


chengyin:
Hi there! Thank you for the translation. This is fantastic! 👍🏼

Some suggestions:

1. Can we use the proper Chinese punctuations instead of the English ones?
2. For the [comment section](https://github.com/1c7/ruby-airbnb#文件类-级别的注释-fileclass-level-comments) I am not sure if the example comments in the code needs to be translated
3. I recommend use 正确 and 错误 instead of 不好 and 好

Let me know what do you think. Glad to help if you have any questions. Once again, thank you for the great work!

1c7:
@chengyin 
1. Sure, no problem, I would replace these English puncations to Chinese
2. I think you are right, let me turn that back to English, comment is not really necessary to translate
3. Sure. 

:D

1c7:
@chengyin  @ljharb  Done.
:D

chengyin:
Thank you!

1c7:
@chengyin  :D

PR #104, opened 2016-08-23
Define preference for raising args over instances

carmi:
Adding this (copied from https://github.com/bbatsov/ruby-style-guide) after seeing this preference internally:

* <a name="exception-class-messages"></a>
  Prefer supplying an exception class and a message as two separate arguments
  to `raise`, instead of an exception instance.
<sup>[[link](#exception-class-messages)]</sup>

  ```Ruby
  # bad
  raise SomeException.new('message')
  # Note that there is no way to do `raise SomeException.new('message'), backtrace`.

  # good
  raise SomeException, 'message'
  # Consistent with `raise SomeException, 'message', backtrace`.
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1212,6 +1212,23 @@ In either case:
     raise MyExplicitError
     ```
 
+
+* <a name="exception-class-messages"></a>
+    Prefer supplying an exception class and a message as two separate arguments
+    to `raise`, instead of an exception instance.
+    <sup>[[link](#exception-class-messages)]</sup>
+
+    ```Ruby
+    # bad
+    raise SomeException.new('message')
+    # Note that there is no way to do `raise SomeException.new('message'), backtrace`.
+
+    # good
+    raise SomeException, 'message'
+    # Consistent with `raise SomeException, 'message', backtrace`.
+    ```
+
+
 * <a name="rescue-as-modifier"></a>Avoid using rescue in its modifier form.
     <sup>[[link](#rescue-as-modifier)]</sup>
 


ljharb:
LGTM, but let's get a few more eyes on it

robotpistol:
seems reasonable to me.  but yeah we should get more eyes on it.

BMorearty:
Seems reasonable to me. Hopefully Rubocop won't enforce it in cases where the constructor takes additional args, though, since Ruby only supports this syntax if the 2nd arg is the message.

carmi:
@BMorearty relating to Rubocop's enforcement:

Right now this cop is disabled for us.

The rubocop config for this is:

```
  Description: Checks the arguments passed to raise/fail.
  StyleGuide: https://github.com/bbatsov/ruby-style-guide#exception-class-messages
  Enabled: false
  EnforcedStyle: exploded
  SupportedStyles:
  - compact
  - exploded
```

So I'm not sure if/when we'd enforce this with Rubocop.

robotpistol:
One way to find out. turn it on locally and run rubocop to find all the instances of it. you might be able to autocorrect them.

carmi:
So I tested this, and yes, if you enable Rubocop with the `EnforcedStyle: exploded` setting enabled it will complain when this style is not followed.

So, I say we merge this, recommend this style, but not yet enforce with Rubocop. Sound good?

BMorearty:
Sounds good to me.

tommydangerous:
LGTM

PR #108, opened 2016-10-15
Update GitHub ruby style guide url

broccolini:
This is publicly accessible if you want to link directly to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This is Airbnb's Ruby Style Guide.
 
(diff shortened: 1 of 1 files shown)


ljharb:
The old version was taken down, and this one is new - I'd prefer to link to both.

pthomas551:
+1

shashankjo (approved):

pariser:
Unfortunately the "live" link to the github ruby style guide is no more. Keeping the link to the internet archive (which still works).

PR #111, opened 2017-03-03
Update options hash example style

kmsun07:
using `reverse_merge!` performs a mutation - since the hash can be passed by reference, it may mutate values on the object even outside the function. Switch to an assignment syntax so that we get a new reference/object and thus do not cause a side effect to the caller

would also be open to 
```
options = options.reverse_merge({
  # the default options in-lined
})
```

@robotpistol


robotpistol:
LGTM

PR #114, opened 2017-04-28
Avoid nested conditionals in favor of early returns.

aagrawal2001:
This PR adds the "early return" style. The idea is to avoid nested conditionals and complex booleans and have functions bail early.

@kmsun07 @robotpistol

### DIFF
--- a/README.md
+++ b/README.md
@@ -797,6 +797,71 @@ In either case:
     end
     ```
 
+### Nested conditionals
+
+* <a name="no-nested-conditionals"></a>
+  Avoid the use of nested conditionals for flow of control. 
+  ([More on this][avoid-else-return-early].) <sup>[[link](#no-nested-conditionals)]</sup>
+  
+  Prefer a guard clause when you can assert invalid data. A guard clause
+  is a conditional statement at the top of a function that returns as soon
+  as it can. 
+  
+  The general principles boil down to:
+  * Return immediately once you know your function cannot do anything more.
+  * Reduce nesting and indentation in the code by returning early. This makes
+  the code easier to read and requires less mental bookkeeping on the part
+  of the reader to keep track of `else` branches.
+  * The core or most important flows should be the least indented.
+
+  ```ruby
+  # bad
+  def compute
+    server = find_server
+    if server
+      client = server.client
+      if client
+        request = client.make_request
+        if request
+           process_request(request)
+        end
+      end
+    end
+  end
+
+  # good
+  def compute
+    server = find_server
+    return unless server
+    client = server.client
+    return unless client
+    request = client.make_request
+    return unless request
+    process_request(request)  
+  end
+  ```
+
+  Prefer `next` in loops instead of conditional blocks.
+
+  ```ruby
+  # bad
+  [0, 1, 2, 3].each do |item|
+    if item > 1
+      puts item
+    end
+  end
+
+  # good
+  [0, 1, 2, 3].each do |item|
+    next unless item > 1
+    puts item
+  end
+  ```
+
+  See also the section "Guard Clause", p68-70 in Beck, Kent.
+  *Implementation Patterns*. Upper Saddle River: Addison-Wesley, 2008, which
+  has inspired some of the content above.
+
 ## Syntax
 
 * <a name="no-for"></a>Never use `for`, unless you know exactly why. Most of the
@@ -1596,6 +1661,7 @@ In either case:
 [google-c++-comments]: https://google.github.io/styleguide/cppguide.html#Comments
 [google-python-comments]: https://google.github.io/styleguide/pyguide.html#Comments
 [ruby-naming-bang]: http://dablog.rubypal.com/2007/8/15/bang-methods-or-danger-will-rubyist
+[avoid-else-return-early]: http://blog.timoxley.com/post/47041269194/avoid-else-return-early
 
 ## Translation
 


kmsun07:
LGTM

aagrawal2001:
@BMorearty

BMorearty:
Is this one guideline? It kind of looks like 2 or 3.

"Avoid nested conditions" seems useful. So does the untangling of complex Booleans, although I'd probably consider that a separate guideline. I'd be in favor of the first, and ok with the second.

Re: `next`, I could take it or leave it. No opinion.

aagrawal2001:
@BMorearty I copy/pasted much of this from https://github.com/bbatsov/ruby-style-guide which ours is based off and many other examples are also copied from there. It combines the  ```next``` guideline there, which makes sense to me since the overall idea here is to bail early. 

I do see the point, however, that untangling booleans is not exactly the same as nested conditionals, but the remedy is the same. So I could move this to a separate section entirely called "Early Returns". Or leave it as-is. Thoughts?

README.md
```diff
@@ -797,6 +797,82 @@ In either case:
     end
     ```
 
+### Nested conditionals
+
+* <a name="no-nested-conditionals"></a>
+  Avoid the use of nested conditionals for flow of control. 
+  ([More on this][avoid-else-return-early].) <sup>[[link](#no-nested-conditionals)]</sup>
```
akuhn: No clear to me why a random blog post is an authoritative source. Our styleguide is authoritative by and in itself, no need to link to external sources. 

If you do feel this matter is controversial enough to require external validation, maybe link to Kent Beck's implementation pattern book, and the section on guard clauses.

See this sample chapter http://ptgmedia.pearsoncmg.com/images/9780321413093/samplechapter/BeckCh_0321413091.pdf

README.md
```diff
@@ -797,6 +797,82 @@ In either case:
...
+
+  Prefer a guard clause when you can assert invalid data. A guard clause
+  is a conditional statement at the top of a function that bails out as
```
akuhn: Do not use jargon, non-native speakers will not understand "bail out"—instead use technical terms like "return"

README.md
```diff
@@ -797,6 +797,82 @@ In either case:
...
+  soon as it can. 
+  
+  The general principles boil down to:
+  * Return as soon as you know your method cannot do any more meaningful work
+  * Reduce indentation by using `if`/`unless`/`return` instead of a top-level 
+  `if`/`else`
+  * Try keep the “meat” of your method at the lowest indentation level.
```
akuhn: Love that! Kent Beck formulates this as "the most important code flow should have the least indentation"

kmsun07:
seems like complex conditionals belongs in `conditional expressions`. maybe we should make a `conditionals` supersection to hold these?

README.md
```diff
@@ -797,6 +797,82 @@ In either case:
...
+
+  ```ruby
+  # bad
+  def compute_thing(thing)
+    if thing[:foo]
+      update_with_bar(thing[:foo])
```
akuhn: Try to make actual examples rather than `foo` and `bar`

For example Kent uses
```
void compute() {
  Server server= getServer();
  if (server != null) {
    Client client= server.getClient();
    if (client != null) {
      Request current= client.getRequest();
      if (current != null) {
        processRequest(current);
      }
    }
  }
}
```

BMorearty:
@aagrawal2001 The complex booleans does seem like a separate guideline to me. It's unrelated to nested conditionals.

I do like early returns as a way—sometimes—to clarify code. But I'm not sure it's right to say it's better in the general case than having a complex conditional. It's funny, looking at the code sample again, the complex conditional is easier for me to understand than the early return.

Plesae take all this as input, but not strong input. I don't feel strongly about any of it.

README.md
```diff
@@ -797,6 +797,82 @@ In either case:
...
+    puts item
+  end
+  ```
+
+  The same technique can be used to untangle complex booleans. 
+  
+  ```ruby
+  # bad
+  def should_do_something?
+    condition_a && (((condition_b || condition_c) && condition_d) || condition_e)
```
akuhn: I don't feel this is a good example for refactoring towards guarding ifs. The result below with its mix of `false unless` and `true if` is just as unreadable. Complex booleans like this should rather assigned partial predicates to intention revealing local variables.
```
def should_do_something?
  user_has_been_ghosted = ((condition_b || condition_c) && condition_d)
  condition_a && (user_has_been_ghosted || condition_e)
end
```

README.md
aagrawal2001: For bang methods, we also cite a blog: http://davidablack.net/dablog.html#2007/8/15/bang-methods-or-danger-will-rubyist 
I cited this one because it did a very good job articulating the rationale behind this. I think it's helpful for people to read it if they want a deeper dive. A second reason is I am copying some of the blog's wording and it seems unfair to do so without citing the author and giving them credit. 

Kent Beck's explanation is really nice too. I can link to that in addition, unless you have any objections?

README.md
aagrawal2001: Hmm.. good point. I like the local variables, but I also want us to return early to avoid doing a bunch of unnecessary computation. A good example is a ```is_valid?``` method which can return ```false``` if a bunch of fields aren't valid. It doesn't need to do ```a_valid && b_valid && c_valid && d_valid && e_valid ```.

Do you feel such an example might be more clear? I'm trying to avoid using actual production code which is making it harder to come with an example.

README.md
aagrawal2001: ok, will change this.

README.md
aagrawal2001: thanks for the suggestion. i will rework this with some more realistic examples.

README.md
akuhn: `&&` and `||` are lazy

aagrawal2001: Right, but we can potentially avoid calculating ```user_has_been_ghosted``` entirely if ```condition_a``` is false.

aagrawal2001:
@akuhn @BMorearty removed the booleans from this for now. Also re-worked the example to be more real-world. PTAL.

akuhn:
LGTM

Issue #74, opened 2016-02-10
Link to rationales?


ljharb:
Perhaps a nice way to mitigate the concerns of those who want the guide to be approachable by non-rubyists, without compromising the terseness that some value, would be to add an explicit "Rationale" link to each rule/guideline that links to the pull request where it was discussed?

That would avoid readers having to hunt through git history to figure out why a rule exists.

Another alternative would be to make a separate `rationales.md` document that was a longer version of the guide, and we could deep link into *that* from the main readme.

Thoughts?

robotpistol:
:+1:

tommydangerous:
@ljharb I love the `rationales.md`

ljharb:
Immediately :-) PRs welcome! :-p

venantius:
I'm a fan of this, especially if we enable rich linking in the short guide to the specific section in the long guide.

BMorearty:
A link to the PR seems like a good, low-cost way to do it. I like that one.

If you do a separate `rationales.md`, I think it'll be out of sync a lot. I wouldn't recommend a "long version" of the guide that repeats all the guidelines, either--then it will be even more out of sync.

Since these are style issues, a lot of the time there's no rationale other than "we wanted to pick a style so we picked this one." :stuck_out_tongue_winking_eye:

kmsun07:
we could always of course add an appendix in the main file, then link down to it from the main rule

ljharb:
Even when the rationale is "the N forms are equivalent and we just like this one better" that's hugely valuable - it informs users which rules they don't have to consider too heavily before deviating from in their own projects.

As long as the rationale is immediately reachable, I will be happy. Requiring users to track down the appropriate PR is super hostile to people not intimately familiar with git/github.

imo this guide shouldn't be primarily or solely aimed at rubyists, it should be aimed at non-rubyists who want to use Ruby (especially newcomers to Ruby that work at Airbnb) - and as such, the rationale is pretty critical to understand as readers are learning the language.

tommydangerous:
I see two great ideas, a separate file for the rationales and the appendix that @kmsun07 proposed.

BMorearty:
As long as "arbitrary consistency" is a valid rationale. I'm not being flippant or sarcastic. For many--I would say most--style rules, that's the reason. Otherwise I would have a hard time justifying, for example, two-space indentation vs. some other number of spaces.

ljharb:
[...]

PR #115, opened 2017-05-02
Add rule discouraging `unless` with comparators

kmsun07:
Prefer `if` with the reverse comparator if available

@robotpistol @aagrawal2001 @BMorearty

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,8 +727,6 @@ In either case:
         ...
       end
     ```
-    
-    
 
 * <a name="unless-with-comparator"></a>Avoid `unless` with comparators if you can use `if` with an opposing comparator.<sup>[[link](#unless-with-comparator)]</sup>
 
--- a/README.md
+++ b/README.md
@@ -727,6 +727,32 @@ In either case:
         ...
       end
     ```
+    
+    
+
+* <a name="unless-with-comparator"></a>Avoid `unless` with comparators if you can use `if` with an opposing comparator.<sup>[[link](#unless-with-comparator)]</sup>
+
+    ```ruby     
+      # bad
+      unless x == 10
+        ...
+      end
+      
+      # good
+      if x != 10
+        ...
+      end
+      
+      # bad
+      unless x < 10
+        ...
+      end
+      
+      # good
+      if x >= 10
+        ...
+      end
+    ```
 
 * <a name="parens-around-conditions"></a>Don't use parentheses around the
     condition of an `if/unless/while`.


kmsun07:
i wanted to add a `blank?`/`present?` example but those come from activesupport. i could use `zero?`/`nonzero?` or something like that, but the examples are cloudy if they're not as relatable

BMorearty:
Seems raisinible.

README.md
```diff
@@ -728,6 +728,30 @@ In either case:
       end
     ```
 
+* <a name="unless-with-comparator"></a>Avoid `unless` with comparators if you can use `if` with an opposing comparator.<sup>[[link](#unless-with-comparator)]</sup>
```
aagrawal2001: What's the rationale for this rule? Can we cite some references here?

aagrawal2001:
```blank?``` and ```present?``` aren't comparators. I think the rule you're trying to maybe write is prefer ```if``` when there's a direct inversion of the ```unless``` form..

**Comment by BMorearty** [comment]
> I think the rule you're trying to maybe write is prefer if when there's a direct inversion of the unless form..

But there's always a direct inversion of the `unless` form, unless I misunderstand what you mean by "direct inversion." That would be the same as saying "never use unless," which I'm strongly against. In some cases `unless` is more readable. That's why it's there.

File: README.md (hunk 1 above)
Comment (BMorearty) [inline README.md]: Do we always need to cite rationales or references? In the past we haven't required this. I don't feel we need to require it every time.

Comment (aagrawal2001) [inline README.md]: I strongly think we should endeavor to do so. I don't really understand why we're making this part of our style guide. I can imagine others might wonder too. We're better off if people understand why they're doing something rather than just having it mandated upon them.

**Comment by aagrawal2001** [comment]
I think Kevin's rule is quite specific and so are the examples - they are all using comparison operators which have a direct opposite operator. An example is a random function like 
``` 
unless object.is_done?
   object.process
end
``` 

Where no inverse function for ```is_done?``` is defined (unlike ```blank``` and ```present?```). Then it becomes unwieldy to write it as ``` if !object.is_done?```

Of course I'm extrapolating. If the rule is indeed to *never* use ```unless`` then that's far too broad.

README.md
BMorearty: Well, we've had this discussion before and landed on the side of keeping it concise instead of wordy. If the rationale for this particular guideline isn't clear, maybe it shouldn't be included or we make an exception and include a very brief rationale here, as have in a few other cases.

aagrawal2001: Ok, I wasn't in that discussion, but a brief rationale is fine. Especially if you link to a source somewhere where it is explained in greater detail. I would like to read the explanation myself frankly.

romanfuchs:
👍

kmsun07:
I think it's generally just `unless x` vs `if !x` where `unless` is preferrable

kmsun07:
but even for that i've seen contention, though mostly i think it's from people who are more accustomed to languages where `unless` doesn't exist

kmsun07:
anyway, if we go that way, i think we should have that in a separate rule

README.md
kmsun07: do we need a rationale? could it be as simple as arbitrarily picking a standard? my rationale is that it's more readable, but that's an eye of the beholder style argument

ljharb:
That's fair; but I think having "don't use unless when blah" in the guide without "here's where it's OK to use unless" is problematic.

romanfuchs:
+1. Most non-native speakers have a strong aversion to `unless`. I always use `if !` because `IF` and `NOT` are standard, formal logical terms, `UNLESS` is not, and hence is slower to parse for a lot of people. (Anyway, I'm ok if some people use it, but definitely don't force me to..)

aagrawal2001:
Can we elaborate and put this part into the rule then: 
I think it's generally just unless x vs if !x where unless is preferrable

And yes it would be super helpful to show examples where unless is preferrable.

kmsun07:
that's a separate rule and should be discussed separately. this rule is self describing already

i can open a PR for that other discussion, but i don't feel super strongly one way or another (in terms of enforcing, that is; i definitely prefer using `unless`) so i won't be pushing it through.

README.md
aagrawal2001: We have to arbitrarily pick a standard if it's purely just about a style choice (like 2 spaces vs. 4) and we cannot find any good objective argument to support it.

Is this one of those cases? Or are we somehow trying to make the code easier to read/understand? Maybe it sounds more natural to the spoken language? There's some discussion here: https://github.com/bbatsov/ruby-style-guide/issues/329

Or maybe the rationale is it's easier to think in terms of ```if``` rather than ```unless``` because the latter is a negation.

ljharb:
I don't think it is a separate rule, and I don't think this should have been merged over multiple people stating they wanted the example.

aagrawal2001:
I have to agree with @ljharb -- it's bad form to merge this when people are in disagreement. We should be in alignment before merging this sort of stuff. It's not that much work to add the clarification that was requested. Also a brief rationale would be super helpful to understand why you are proposing this rule. If we really want people to adopt the rule, this extra effort will pay off. 

Personally, I don't see a huge difference between ```if x != 0``` and ```unless x == 0```, but I realize this is getting into contentious territory.

kmsun07:
you are free to PR the separate rule if you like. you can see literally every other rule in this section has no rationale (short of one that says "it's just not worth it"), nor a positive usage of an unrelated concept

https://github.com/airbnb/ruby#multiline-if-then - should we explain when we CAN use `then`?
https://github.com/airbnb/ruby#multiline-while-until - should we give arbitrary examples of using `do`?
https://github.com/airbnb/ruby#no-unless-with-else and https://github.com/airbnb/ruby#unless-with-multiple-conditions - should we repeat this example that you're asking for in these rules as well?

the rule is perfectly clear on its own. we should not feel the need to justify each keyword in every rule that tangentially touches on it

@ljharb you yourself approved this PR
screen shot 2017-05-02 at 9 17 49 pm

tommydangerous:
I LOVE THIS! THANK YOU.

aagrawal2001:
Just because existing rules lack an explanation does not justify continuing to do so. Several rules do have an explanation. I'm not sure why there's resistance to adding explanations or clarifications, especially when it's something very simple? Some rules have a link explaining in great detail the motivation behind the rule The idea is not to just enforce these rules dogmatically. We'll get much better adoption if we can justify the rationale. 

As for the other rules you're listing, feel free to improve them, but I don't see how that can serve as justification for not clarifying your rule.

ljharb:
It's not clear on its own.

A PR approval in good faith does not mean you can merge through objections from multiple collaborators. If you prefer me to block PRs indefinitely pending my concerns being addressed, I can, but that's generally not the way things are done - neither is merging a PR before addressing all concerns on the thread.

kmsun07:
@ljharb my understand is actually that's exactly what the approve/request changes feature is for - for differentiating between feedback and action-required feedback. if an approval isn't an actual approval, what's the point?

as for the concerns, i feel like they were addressed, repeatedly. i cited many examples of no rationales provided, and brian highlighted that omitting rationales is our convention.

@aagrawal2001 it was an active decision to not include rationales. we in fact created a separate (poorly maintained?) file https://github.com/airbnb/ruby/blob/master/rationales.md for the purpose of composing rationales without bloating this guide.

for both of you - can you explain why adding a positive example of `unless` has anything to do with this rule, which solely deals in cases where `unless` should be avoided?

kmsun07:
for the discussion on separating rationales, see https://github.com/airbnb/ruby/issues/74

if you assert that this rule requires a rationale in that file, i can add that. it does generally seem like a much more innocent rule than many unjustified rules, so i'm not quite understanding why this is where the buck suddenly stops

ljharb:
A rule that describes where X should be avoided, but leaves doubt as to when X would ever be appropriate, should also make it clear as to when X should *not* be avoided.

kmsun07:
is it not fair to say it is not necessarily avoided if you aren't using an invertable comparator?

kmsun07:
and once again, i would refer you to every single other rule under 'conditional keywords' aside from https://github.com/airbnb/ruby#only-simple-if-unless , which follow the same pattern. we shouldn't have to repeat `unless x` as an example for every single rule that references `unless` (of which there are at least 3)

aagrawal2001:
@kmsun07 both brian and I agreed that a short explanation would suffice, but you ignored it and decided to just impose a style. This isn't good because we end up with a bunch of people just blindly following the style without knowing why. Once a developer understands the why and is convinced of it, the enforcement becomes automatic and they spread the knowledge to others. Without it, you have reluctant conformance to some seemingly arbitrary rule created by a select few.

There is also precedent of linking to more detailed explanations (such as the one for freezing constants where you read about freezing strings -- this is an example benefit of such a link). 

This should be done in the style guide itself given how short it can be, not another separate file. 

A positive example of unless is useful because at least 3 people were confused by your rule and started thinking perhaps it means we're banning unless. This illustrates the fact that the rule is unclear. You might think it is clear because you're the author, but you are writing for people to read and understand. 

I'm really not convinced by using other rules as examples as you keep doing. Frankly, many of these are simply copy/pasted from the other guide. If we read your rule it says:
"if you can use if with an opposing comparator". Yet you give no examples where the if cannot be used with an opposing comparator. You're also vague about "comparator". Is it a comparison operator? Or does it include blank? and present? like you alluded to in your comments. This is what is creating some of the confusion. It's just really puzzling to see your reluctance to add clarification...

kmsun07:
* the rationale is arbitrary, there is no real reason - standardization for its own sake is a real thing (and is often the entire point of style guides). different people find different things readable. if you prefer the exact opposite of the rule (always use `unless` over `if`), I'd disagree, but I'd also prefer that over no rule at all

* once again, a generic positive example of `unless` is meaningless because this example does not have anything to do with the general case of `unless`. it is specifically concerned about the usage of `unless` alongside a reversible comparison operator. do i need to provide a generic negative example of `if` as well?

* a comparator is a real life concept. i guess i could change it to the slightly more official name of 'comparison operator'? we shouldn't be in the business of glossarifying generic programming terms.

if it really helps, i can add that `unless x === y` is ok (which itself surprises me that there is no `!==` - i had previously assumed they were all invertible)

generally, i'm reluctant to add rationale text because we have actively decided to avoid doing so in the style guide. I am reluctant to add a general example of `unless` without a comparator because it is not directly related to the rule. now that i'm aware of `===`, i'm happy to add that as an 'acceptable' example

kmsun07:
opened https://github.com/airbnb/ruby/pull/117 for the new example and text correction of comparator to comparison operator (which is indeed far more googlable)

aagrawal2001:
* Any time you can provide a rationale, you are making adoption easier and removing subjectivity. In your case, I think the argument is you find it easier to read "if" because it's positive whereas "unless" is a negative and requires some mental work to reverse the condition. That's what the message thread was suggesting. The second argument I've heard is non-native speakers might struggle with unless (which is more of a case to ban it, but that's overly broad). If it truly is just totally arbitrary, then what's the criteria for putting it into a style guide? What if you feel one way, and another engineer feels that sometimes one choice might be better than the other depending on the exact thing being checked so no rule should be in place? Then we are stuck. These are some of the dangers of choosing the arbitrary style path. 

It's for this reason that when I put in a rule, I try very hard to justify it (concisely) and link to other sources as well. 

* A comparator is generally thought of as a comparison function, say x.equals?(y) is a comparator. Better to be very clear you are referring to Ruby comparison operators. And as we became more specific we found something new and surprising, which is good. It would certainly be good to add that otherwise I'm left wondering as a reader when an opposing operator exists and when it doesn't.

* And there's also the case of the blank?/present? -- those are not comparison operators, but the rule applies there too. You mentioned that in your comments.. We seem to enforce that with Rubocop actually but it's not in our style guide.. So maybe a better generic rule is about when to use unless vs. if.

BMorearty:
@aagrawal2001 If you're including me in the 3 people who "started thinking perhaps it means we're banning unless," please don't. I was responding to one of your comments and asking what you meant, no what the original guideline meant. It was clear to me the original guideline was clearly not doing that. I don't see how it could be interpreted as banning `unless`.

As has been stated many times in this discussion, we already had a discussion in the past about rationales and we made a decision. And it wasn't so long ago that it's time now to revisit that decision. (My personal notes indicate it was 15 months ago.) It's exhausting and unproductive to continually revisit decisions that were made not very long ago. "Disagree and commit" is the right attitude for such decisions. I understand you didn't know about it at first, but once you were informed of it, early in the discussion, it would be appropriate to move on.

I did agree that a short rationale would suffice for this guideline. I said it in the spirit of compromise because the rationale for this guideline was unclear to you. I was not insisting on it, just trying to mediate between the two of you. And I want to be clear, I do not support revisiting the decision on rationales for all guidelines at this time. If you do want to revisit it, please don't change the rules on the fly in a single PR from someone. Feel free to ask for a rationale for the PR in question, but this is not the place to say we need rationales for all guidelines going forward.

aagrawal2001:
@BMorearty sorry you are frustrated. I feel 15 months is plenty of time. I also feel it's perfectly appropriate to bring up the general question in a PR because it's in context as opposed to some general discussion at a later point without context. We don't have to change rules on the fly, but we shouldn't shut down discussion like this. 
Once again, sorry to frustrate you - not my intent here. And I do appreciate the spirit of compromise.

arianf:
Is there a cop I can enable to get this rule, I would love to enforce this in our projects

PR #116, opened 2017-05-03
Freeze constants.

aagrawal2001:
@kmsun07 @BMorearty @robotpistol

### DIFF
--- a/README.md
+++ b/README.md
@@ -1024,6 +1024,37 @@ In either case:
        an attribute when `self` is an ActiveRecord model: `self.guest = user`.
     3. Referencing the current instance's class: `self.class`.
 
+* <a name="freeze-constants"></a>When defining arrays and hashes meant to be
+    constants, make sure to call `freeze` on them. ([More on this][ruby-freeze].)
+    <sup>[[link](#freeze-constants)]</sup>
+
+    The reason is that Ruby constants are actually mutable. Calling `freeze`
+    ensures they are not mutated and are therefore truly constant and 
+    attempting to modify them will raise an exception. 
+
+    ```ruby
+    class COLOR
+      # Each color
+      RED = 0
+      BLUE = 1
+      GREEN = 2
+      
+      # bad
+      ALL_COLORS = [
+        RED,
+        BLUE,
+        GREEN,
+      ]
+
+      # good
+      ALL_COLORS = [
+        RED,
+        BLUE,
+        GREEN,
+      ].freeze
+    end
+    ```
+
 ## Naming
 
 * <a name="snake-case"></a>Use `snake_case` for methods and variables.
@@ -1596,6 +1627,7 @@ In either case:
 [google-c++-comments]: https://google.github.io/styleguide/cppguide.html#Comments
 [google-python-comments]: https://google.github.io/styleguide/pyguide.html#Comments
 [ruby-naming-bang]: http://dablog.rubypal.com/2007/8/15/bang-methods-or-danger-will-rubyist
+[ruby-freeze]: http://blog.honeybadger.io/when-to-use-freeze-and-frozen-in-ruby/
 
 ## Translation
 


README.md
```diff
@@ -1113,6 +1113,37 @@ In either case:
...
+    <sup>[[link](#freeze-constants)]</sup>
+
+    The reason is that Ruby constants are actually mutable. Calling `freeze`
+    ensures they are not mutated and are therefore truly constant and 
+    attempting to modify them will raise an exception. 
+
+    ```ruby
+    class COLOR
+      # Each color
+      RED = 0
```
kmsun07: as a bonus sub-example, would we want to switch the examples to something like `RED = 'red'.freeze`?

README.md
```diff
@@ -1113,6 +1113,37 @@ In either case:
...
+* <a name="freeze-constants"></a>When defining arrays and hashes meant to be
+    constants, make sure to call `freeze` on them. ([More on this][ruby-freeze].)
+    <sup>[[link](#freeze-constants)]</sup>
+
+    The reason is that Ruby constants are actually mutable. Calling `freeze`
+    ensures they are not mutated and are therefore truly constant and 
+    attempting to modify them will raise an exception. 
+
+    ```ruby
+    class COLOR
```
kmsun07: this example violates the `CamelCase` class convention, maybe rename this to `Color`?

aagrawal2001: Very good catch. I will fix this! Thanks!

README.md
aagrawal2001: I was first going to write it that way after reading the link explaining the efficiency gain, but then I decided to scope down the rule to just arrays and hashes because we totally don't do this in our code base at all. 

Totally open to expanding this rule to strings, and in fact any type really (even RED = 0.freeze is ideal because we don't want some other code to re-assign RED), but again none of our code does it, so it felt a bit hypocritical. 

Thoughts?

aagrawal2001: I think we'd have to start actually doing this though and it might feel a bit much at first :-)

kmsun07: some types are already frozen by default, like numbers and symbols.
```
2.4.0 :001 > 0.frozen?
 => true
2.4.0 :002 > :test.frozen?
 => true
```
and freezing doesn't actually prevent reassignment. i'm not sure if there is any way to prevent reassignment at all, short of choosing a paradigm, I think
```
2.4.0 :003 > x = 'a'.freeze
 => "a"
2.4.0 :004 > x = 'b'.freeze
 => "b"
```

aagrawal2001: I meant to say mutate, not re-assign. Strings can be accidentally mutated, just like hashes and arrays, but we rarely do that in our code. So that's why it felt weird adding that rule in. But, I can do it if we feel this is practical/useful to do going forward.

(Good to know that numbers are immutable, so at least we don't have to worry about those)

kmsun07: i think we do strive to freeze them as well, we just aren't as disciplined about it because it bites us less often

README.md
BMorearty: @aagrawal2001 It looks like this didn't get fixed.

aagrawal2001: @BMorearty yes I know and I will fix it shortly. Rest assured, I did not forget.

aagrawal2001: https://github.com/airbnb/ruby/pull/118

README.md
aagrawal2001: Searching in the code, it looks like we're mainly doing it where performance has bitten us, and not as a general practice. 

For a lot of the missing ones, it feels like symbols would be better anyway. So perhaps we can have some guidelines around when to prefer symbols and then add that string constants should be frozen. Or it could be a separate guideline linked from this one too since it is a thing of its own.

robotpistol: There is a rubocop rule to enforce this.

kmsun07: symbols are touchy because if constants get used as enums and then written to the DB (as we commonly do), they may not come back as symbols

ljharb: While in those cases we could override accessors to symbolize them, that's definitely a footgun to be aware of.

aagrawal2001: Yeah I guess we seem to use integers and then have a mapping from those to the strings. So we're not using symbols at all. Maybe that's ok.

PR #117, opened 2017-05-03
Update unless with comparator rule with unreversible example

kmsun07:
also rename comparator to the more official name, comparison operator

i still maintain there is no rationale here and thus no need to force one. previous discussion is once again at https://github.com/airbnb/ruby/issues/74

please do not approve this PR unless you are OK with it being merged as is (that is, you would consider it an improvement over the currently live version), and please request changes if you have what you consider to be showstopper suggestions

@ljharb @aagrawal2001

### DIFF
--- a/README.md
+++ b/README.md
@@ -728,7 +728,7 @@ In either case:
       end
     ```
 
-* <a name="unless-with-comparator"></a>Avoid `unless` with comparators if you can use `if` with an opposing comparator.<sup>[[link](#unless-with-comparator)]</sup>
+* <a name="unless-with-comparison-operator"></a>Avoid `unless` with comparison operators if you can use `if` with an opposing comparison operator.<sup>[[link](#unless-with-comparison-operator)]</sup>
 
     ```ruby     
       # bad
@@ -748,6 +748,11 @@ In either case:
       
       # good
       if x >= 10
+        ...
+      end
+      
+      # ok
+      unless x === 10
         ...
       end
     ```


README.md
```diff
@@ -728,7 +728,7 @@ In either case:
       end
     ```
 
-* <a name="unless-with-comparator"></a>Avoid `unless` with comparators if you can use `if` with an opposing comparator.<sup>[[link](#unless-with-comparator)]</sup>
+* <a name="unless-with-comparison-operator"></a>Avoid `unless` with comparison operators if you can use `if` with an opposing comparison operator.<sup>[[link](#unless-with-comparison-operator)]</sup>
```
ljharb: i don't object to the name "comparator", fwiw.

kmsun07: comparison operator is more googlable (searching for ruby comparator gives comparison operator in the titles) so i think it's worth changing to avoid having to explain it here

kmsun07:
merging as i would consider this a strict improvement (and it seems agreed) - happy to continue further conversation about rationales, etc.

kmsun07:
it turns out that `===` means nothing like what i thought it meant http://stackoverflow.com/questions/4467538/what-does-the-operator-do-in-ruby . it still does seem like it qualifies as a comparison operator though? i suppose in reality, there's never actually such thing as a comparison operator, and it's just generally conventions that are implemented on relevant objects

README.md
ljharb: fair point. i do think of "function that compares two things" when i think "comparator"

aagrawal2001:
It does feel like an operator, but I've rarely if ever seen it used.

kmsun07:
to be fair, unlike the other things we normally think of as comparison operators, `===` is actually implemented on `Object` itself so arguably it's the most valid one! http://ruby-doc.org/core-2.4.1/Object.html#method-i-3D-3D-3D

PR #119, opened 2017-05-03
Add strings to list of freeze recommended types.

aagrawal2001:
@kmsun07  adding strings to freeze recommended types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ Airbnb also maintains a [JavaScript Style Guide][airbnb-javascript].
 
 ### Newlines
 
-* <a name="multiline-if-newline"></a>Add a new line after `if` conditions span
+* <a name="multiline-if-newline"></a>Add a new line after `if` conditions spanning
     multiple lines to help differentiate between the conditions and the body.
     <sup>[[link](#multiline-if-newline)]</sup>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1118,20 +1118,21 @@ In either case:
        an attribute when `self` is an ActiveRecord model: `self.guest = user`.
     3. Referencing the current instance's class: `self.class`.
 
-* <a name="freeze-constants"></a>When defining arrays and hashes meant to be
-    constants, make sure to call `freeze` on them. ([More on this][ruby-freeze].)
-    <sup>[[link](#freeze-constants)]</sup>
+* <a name="freeze-constants"></a>When defining strings, arrays or hashes meant to
```
ljharb: Oxford comma after "arrays" please.

Also, isn't it *any* mutable type? A Set should be frozen too, for example.

kmsun07: oxford comma?

ljharb: jinx

README.md
```diff
@@ -1141,6 +1142,11 @@ In either case:
       ]
 
       # good
```
kmsun07: is this block weird to read since it's all inside the same `class` and we're redeclaring (and reassigning) the same constants? what if did `RED`as bad and `GREEN` as good or something like that? or alternately, use two different `class Color` blocks and do the `bad` / `good` declaration out there?

aagrawal2001: I was looking at one of our other class blocks below and it also does the re-declaration thing, so that's why I did it this way. I can make it a different block if we feel that'll be cleaner.

README.md
aagrawal2001: Good point. Should we make the rule generic then to include all mutable types? Maybe say something like:

When defining an object of any mutable type meant to be a constant, make sure to call `freeze` on it. Common examples are strings, arrays, and hashes. 

(Rubocop only seems to allow enforcement of these three btw).

aagrawal2001:
@ljharb @kmsun07 take a look now.

README.md
ljharb: Rubocop won't always be able to enforce things in the guide; the guide will often be more expansive than rubocop is capable of handling.

Issue #101, opened 2016-06-25
Prefer map over collect ?


akash2504:
This style guide says that we should prefer map over collect. but this link (http://rubyinrails.com/2014/01/25/ruby-difference-between-collect-and-map/) clearly explains that in ruby there is no difference between map and collect. Then why to prefer map?
Am i missing something?

ljharb:
Preferring one over the other helps maintain consistency - in the common case of there being multiple aliases for the same function, it's a best practice to select one variant, and require that be the one used.

In regard to this specific case, imo `map` is a more common idiom for "transform one array into another", and as such, that name is less confusing than the much rarer "collect".

tommydangerous:
+1 to @ljharb

akash2504:
We should mention this in the style guide. Having the statement Prefer map over collect gives a sense that the implementation of two is different and former is better over latter. Isn't it?

ljharb:
I both agree that more explanation would be better, and disagree that the current statement implies anything of the sort whatsoever.

harry-wood:
Also in the ruby style guide: https://rubystyle.guide/#map-find-select-reduce-size 

_"Prefer `map` over `collect`. ... This is not a hard requirement; if the use of the alias enhances readability, it’s ok to use it "_

PR #120, opened 2017-05-05
Additional detail regarding the use of map over collect

bob-nomann:
Added additional detail to style guide regarding the use of map and collect. 
Closes airbnb/ruby#101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1390,7 +1390,7 @@ In either case:
 ## Collections
 
 * <a name="map-over-collect"></a>Prefer `map` over
-    `collect`.<sup>[[link](#map-over-collect)]</sup>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1390,7 +1390,7 @@ In either case:
 ## Collections
 
 * <a name="map-over-collect"></a>Prefer `map` over
-    `collect`.<sup>[[link](#map-over-collect)]</sup>
+    `collect` as it's a more recognized convention among programmers.<sup>[[link](#map-over-collect)]</sup>
```
akuhn: Shouldn't we add the same comment for `reduce` then?

ljharb: Yes, we should use reduce over inject for the same reason imo.

shashankjo (approved):

PR #121, opened 2017-05-15
Update example for throwaway variable

kmsun07:
The example cites a function that is not only opaque to most readers, but doesn't even exist in airbnb's codebase anymore. Thus, the interpretation may be lost, as nobody knows what the expected return value(s) of the function is. Replace it with standard ruby function(s)

@robotpistol

### DIFF
--- a/README.md
+++ b/README.md
@@ -1194,9 +1194,12 @@ In either case:
     <sup>[[link](#throwaway-variables)]</sup>
 
     ```ruby
-    payment, _ = Payment.complete_paypal_payment!(params[:token],
-                                                  native_currency,
-                                                  created_at)
+    {
+      a: 1,
+      b: 2,
+    }.each do |key, _|
+      puts key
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1194,9 +1194,12 @@ In either case:
     <sup>[[link](#throwaway-variables)]</sup>
 
     ```ruby
-    payment, _ = Payment.complete_paypal_payment!(params[:token],
-                                                  native_currency,
-                                                  created_at)
+    {
+      a: 1,
+      b: 2,
+    }.each do |key, _|
```
akuhn: Feels like `each_key` should be the preferred choice in this example. Maybe use `split` as an example?
```
major, minor, _ = version_number.split('.')
```

kmsun07: good idea

kmsun07:
@akuhn updated

PR #123, opened 2017-05-17
Recommend named arguments for default arguments

kmsun07:
as an alternative to options hash.

clarify the rule to call out positional arguments only - default named arguments are encouraged.

@robotpistol

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,12 +514,17 @@ Thus when you create a TODO, it is almost always your name that is given.
      end
      ```
 
-* <a name="no-default-args"></a>Do not use default arguments. Use an options
+* <a name="no-default-args"></a>Do not use default positional arguments. Use named arguments or an options
     hash instead.<sup>[[link](#no-default-args)]</sup>
 
     ```ruby
     # bad
     def obliterate(things, gently = true, except = [], at = Time.now)
+      ...
+    end
+
+    # good
+    def obliterate(things, gently: true, except: [], at: Time.now)
       ...
     end
 
--- a/README.md
+++ b/README.md
@@ -514,7 +514,8 @@ Thus when you create a TODO, it is almost always your name that is given.
      end
      ```
 
-* <a name="no-default-args"></a>Do not use default positional arguments. Use named arguments or an options
+* <a name="no-default-args"></a>Do not use default positional arguments. 
+    Use keyword arguments (if available - in Ruby 2.0 or later) or an options
     hash instead.<sup>[[link](#no-default-args)]</sup>
 
     ```ruby


README.md
```diff
@@ -514,7 +514,7 @@ Thus when you create a TODO, it is almost always your name that is given.
      end
      ```
 
-* <a name="no-default-args"></a>Do not use default arguments. Use an options
+* <a name="no-default-args"></a>Do not use default positional arguments. Use named arguments or an options
```
ljharb: Is it worth pointing out the Ruby version restrictions, for gem authors where it tends to matter?

kmsun07: sure - add an `(if available in your version)` maybe?

akuhn: The official name is "keyword arguments," they were introduced with Ruby 2.0 and [performance issues with them were fixed in Ruby 2.2](https://bugs.ruby-lang.org/issues/10440), not sure starting from which version our guide should thus recommend their use.

ljharb: Let's say 2.2 and up only?

kmsun07: this is interesting - I believe we (airbnb) are actually on ruby `2.1.8`. does this mean we should NOT be using keyword arguments?

kmsun07: FWIW, per that link, the performance hit is comparable to using `options` hash (since the issues it cites are around the implied construction/consumption of a `Hash`). If `options` hash is indeed a blessed pattern from a readability/extensibility > performance perspective, should we just recommend keyword arguments from the moment of availability?

aagrawal2001: What is the reason we want to allow default keyword args, but not default positional args? I think one reason behind discouraging default args was to remove the obscurity around them and make it explicit from a caller's standpoint what is being invoked.

ljharb: @kmsun07 i agree that if named args in ruby 2+ are equivalent in perf to an options hash, and there's no other issues, than "use in ruby 2+" is fine with me. if there's any other quirks or caveats, then i would say that yes, we shouldn't use them at all prior to ruby 2.2 (which i assume fixes those quirks).

kmsun07: @aagrawal2001 , i think a major concern with default positional args is the inability to properly have multiple, as to specify the latter ones, you must provide values for the earlier ones. keyword arguments are not subject to this behavior.

positional arguments (optional/defaulted or not) generally being obscure is another question

kmsun07:
updated

README.md
aagrawal2001: @kmsun07 so you mean default positional args must appear at the end of the function signature, but default keyword args can appear anywhere? Like this:
``` def obliterate(gently:, at: Time.now, another_arg:) ```
but we can't do this:
``` def obliterate(gently, at=Time.now, another_arg) ``` ?

I think this only applies to Ruby 2.1 and beyond:
https://robots.thoughtbot.com/ruby-2-keyword-arguments

Why is it bad though? Isn't the main reason that it creates obscurity from an ordering standpoint?

Comment (kmsun07) [inline README.md]: I think it's the inability to provide later values without providing the earlier ones. if we consider the `bad` example
```
def obliterate(things, gently = true, except = [], at = Time.now)
  ...
end
```
if someone wants to call `obliterate` with a custom `at` value, they must ALSO provide values for `gently` and `except`, even if they strictly want to use the default values. if we use the `good` signature, the caller will only specify the exact optional arguments it wants to provide.

i personally also agree that ordering is a problem for a high number of arguments // unintuitive arguments, but I also feel that discussion is tangential to optionality, which provides this sort of weird hard requirement

Comment (aagrawal2001) [inline README.md]: Ok, that makes sense. Might be worthwhile adding this very explanation in there in case others ever wonder about it.

File: README.md [hunk 2]
```diff
@@ -514,7 +514,8 @@ Thus when you create a TODO, it is almost always your name that is given.
      end
      ```
 
-* <a name="no-default-args"></a>Do not use default arguments. Use an options
+* <a name="no-default-args"></a>Do not use default positional arguments. 
+    Use keyword arguments (if available - in Ruby 2.0 or later) or an options
```
aagrawal2001: note that according to https://robots.thoughtbot.com/ruby-2-keyword-arguments
default keyword args are available only in Ruby 2.1+, not Ruby 2.0

kmsun07: i think it's actually the other way around - default keyword args were added in `2.0`, but REQUIRED keyword args were added in `2.1`

tommydangerous:
love it

README.md
aagrawal2001: yes that sounds right to me.

PR #147, opened 2018-10-30
Bump rubocop-rspec to latest, rubocop to 0.58

kevinmsun:
I'm aware this drops support for ruby 2.1, so it might be a non-starter for airbnb. is there an intent to iterate on this gem on the current version and/or to advance the baseline ruby version? @robotpistol @ljharb

### DIFF
--- a/rubocop-airbnb/lib/rubocop/airbnb/version.rb
+++ b/rubocop-airbnb/lib/rubocop/airbnb/version.rb
@@ -3,6 +3,6 @@
 module RuboCop
   module Airbnb
     # Version information for the the Airbnb RuboCop plugin.
-    VERSION = '1.6.0'.freeze
+    VERSION = '2.0.0'.freeze
   end
 end
--- a/rubocop-airbnb/lib/rubocop/airbnb/version.rb
+++ b/rubocop-airbnb/lib/rubocop/airbnb/version.rb
@@ -3,6 +3,6 @@
 module RuboCop
   module Airbnb
     # Version information for the the Airbnb RuboCop plugin.
-    VERSION = '1.5.0'.freeze
+    VERSION = '1.6.0'.freeze
   end
 end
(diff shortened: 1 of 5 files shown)


ljharb:
It’s certainly semver-major; we’d only want to do this if we’ve dropped all our ruby 1.9 apps internally.

rubocop-airbnb/lib/rubocop/airbnb/version.rb
```diff
@@ -3,6 +3,6 @@
 module RuboCop
   module Airbnb
     # Version information for the the Airbnb RuboCop plugin.
-    VERSION = '1.5.0'.freeze
+    VERSION = '1.6.0'.freeze
```
ljharb: iow this would be v2, if it’s dropping support for a ruby version.

robotpistol:
@ljharb We already are running a version of rubocop that dropped support for ruby 1.9 I upgraded our internal stuff after we got monorail off of 1.9.3.  I also agree we should probably do a major version bump here since it does drop support for ruby 2.1.

ljharb:
If that’s the case, then let’s go ahead with this major update.

kevinmsun:
makes sense. i was thrown off by rubocop itself not bumping a major version

kevinmsun:
The Travis CI still runs a ruby 2.1 version of the tests, though

ljharb:
Then we should remove that from travis.yml in this PR :-)

and yeah it’s a bummer how often the ruby community violates semver :-/ but we can still adhere to it :-)

ljharb:
(altho actually, in this case, they didn’t - for versions < v1, the middle identifier is “major”)

robotpistol:
One thing to note for FactoryBot is that newer versions of factory bot no longer differentiate between dynamic and static declarations and require blocks accross the board.

kevinmsun:
i was struggling to find `travis.yml`, but then realized i was only looking inside `rubocop-airbnb`

PR #148, opened 2018-11-09
Attempt to fix list indentation

kevinmsun:
This has been bothering me for a while - https://stackoverflow.com/editing-help#advanced-lists suggests 4 spaces, so trying it out

@ljharb @robotpistol


ljharb:
We should probably make the TOC be autogenerated by a tool too :-)

PR #150, opened 2019-01-06
Update to work on Ruby 2.6 - new versions of rubocop

mcamara:
I know this can't be merged into the master branch because of Airbnb main app restrictions (I think I've read it's using Ruby 2.1 and the latest version of rubocop just accepts 2.2+) but I think having an updated PR that works on Ruby 2.6 can be useful for people that have upgraded their version of Ruby and they still want to use your rubocop ruleset.

Don't hesitate to close this PR if you think this is not appropriate or useful.

Thanks for your work!

### DIFF
--- a/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
+++ b/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
@@ -14,7 +14,7 @@ module RuboCop
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        Util::ASGN_NODES.each do |type|
+        AST::Node::ASSIGNMENTS.each do |type|
           define_method("on_#{type}") do |node|
             enforce_violation(node)
           end
--- a/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
+++ b/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
@@ -14,7 +14,10 @@ module RuboCop
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        AST::Node::ASSIGNMENTS.each do |type|
+        rubocop_version = Gem.loaded_specs.fetch('rubocop', nil)&.version.to_s
+        assignments = rubocop_version.to_f >= 0.59 ? AST::Node::ASSIGNMENTS : Util::ASGN_NODES
+
+        assignments.each do |type|
           define_method("on_#{type}") do |node|
             enforce_violation(node)
           end
--- a/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
+++ b/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
@@ -14,10 +14,13 @@ module RuboCop
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        rubocop_version = Gem.loaded_specs.fetch('rubocop', nil)&.version.to_s
-        assignments = rubocop_version.to_f >= 0.59 ? AST::Node::ASSIGNMENTS : Util::ASGN_NODES
+        assignments = if Gem.loaded_specs['rubocop'].version < Gem::Version.new('0.59.0')
+          RuboCop::Util::ASGN_NODES
+        else
+          RuboCop::AST::Node::ASSIGNMENTS
+        end
 
-        assignments.each do |type|
+         assignments.each do |type|
           define_method("on_#{type}") do |node|
             enforce_violation(node)
           end
--- a/rubocop-airbnb/config/rubocop-layout.yml
+++ b/rubocop-airbnb/config/rubocop-layout.yml
@@ -227,7 +227,7 @@ Layout/FirstMethodParameterLineBreak:
   Enabled: false
 
 # Supports --auto-correct
-Layout/FirstParameterIndentation:
+Layout/IndentFirstArgument:
   Description: Checks the indentation of the first parameter in a method call.
   Enabled: true
   EnforcedStyle: consistent
@@ -238,7 +238,7 @@ Layout/FirstParameterIndentation:
   - special_for_inner_method_call_in_parentheses
 
 # Supports --auto-correct
-Layout/IndentArray:
+Layout/IndentFirstArrayElement:
   Description: Checks the indentation of the first element in an array literal.
   Enabled: true
   EnforcedStyle: consistent
@@ -251,7 +251,7 @@ Layout/IndentAssignment:
   IndentationWidth:
 
 # Supports --auto-correct
-Layout/IndentHash:
+Layout/IndentFirstHashElement:
   Description: Checks the indentation of the first key in a hash literal.
   Enabled: true
   EnforcedStyle: consistent
--- a/rubocop-airbnb/config/rubocop-layout.yml
+++ b/rubocop-airbnb/config/rubocop-layout.yml
@@ -474,11 +474,6 @@ Layout/SpaceInsideBlockBraces:
   EnforcedStyleForEmptyBraces: no_space
   SpaceBeforeBlockParameters: true
 
-Layout/SpaceInsideParens:
-  Description: 'No spaces after ( or before ).'
-  StyleGuide: '#spaces-braces'
-  Enabled: true
-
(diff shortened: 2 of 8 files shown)


rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
```diff
@@ -14,7 +14,7 @@ def enforce_violation(node)
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        Util::ASGN_NODES.each do |type|
+        AST::Node::ASSIGNMENTS.each do |type|
```
ljharb: Is there a way we could make this change such that it uses whichever is available, so that it could work in both 2.1 and 2.6?

mcamara: Good idea! I'll push a solution soon

rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
```diff
@@ -14,7 +14,10 @@ def enforce_violation(node)
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        Util::ASGN_NODES.each do |type|
+        rubocop_version = Gem.loaded_specs.fetch('rubocop', nil)&.version.to_s
```
mcamara: Another way to load assignments would be: (but I think it's too aggressive)
```
begin
  assignments = RuboCop::Util::ASGN_NODES
rescue NameError
  assignments = RuboCop::AST::Node::ASSIGNMENTS
end
```

What do you think?

ljharb: I’m not actually sure what would be a better ruby idiom. perhaps @robotpistol or @BMorearty has thoughts?

mcamara: I should change the safe navigation operator to `try` but I'll wait for their response.
Thanks a lot @ljharb !

tomprats:
I was trying to use this PR since I started a project on ruby 2.6 and still wanted Airbnb's base of rules, but I ran into some trouble. I don't know if it's actually related to the changes here but Rubocop internals are out of current set of knowledge.

I'm using Bundler 2.0.1 and using a glob in my Gemfile since the gem is nested

`gem "rubocop-airbnb", github: "mcamara/ruby", glob: "rubocop-airbnb/*.gemspec"`

When I run `rubocop` I get:

```
cannot load such file -- rubocop-airbnb
/Users/tom/.rvm/rubies/ruby-2.6.0/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/Users/tom/.rvm/rubies/ruby-2.6.0/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:15:in `block in resolve_requires'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:11:in `each'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:11:in `resolve_requires'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader.rb:44:in `load_file'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:140:in `block in base_configs'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:139:in `map'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:139:in `base_configs'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader_resolver.rb:22:in `resolve_inheritance'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader.rb:49:in `load_file'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_loader.rb:82:in `configuration_from_file'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/config_store.rb:44:in `for'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/cli.rb:206:in `apply_default_formatter'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/lib/rubocop/cli.rb:46:in `run'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/exe/rubocop:13:in `block in <top (required)>'
/Users/tom/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/benchmark.rb:308:in `realtime'
/Users/tom/.rvm/gems/ruby-2.6.0/gems/rubocop-0.62.0/exe/rubocop:12:in `<top (required)>'
/Users/tom/.rvm/gems/ruby-2.6.0/bin/rubocop:23:in `load'
/Users/tom/.rvm/gems/ruby-2.6.0/bin/rubocop:23:in `<main>'
```

Potentially relevant configs:

```
# .rubocop_airbnb.yml
require:
  - rubocop-airbnb
```

```
# .rubocop.yml
inherit_from:
  - .rubocop_airbnb.yml
  - .rubocop_todo.yml

inherit_mode:
  merge:
    - Exclude

AllCops:
  Exclude:
    - "db/schema.rb"

...
```

tomprats:
After looking into it further, it looks like I just need to do `bundle exec rubocop` although I haven't been able to figure out why yet. Sorry to bother

robotpistol:
@tomprats You need to `bundle exec rubocop` most likely because `rubocop-airbnb` is not installed to your system wide gems.

If you install the correct version of rubocop-airbnb globally then you should be able to just run `rubocop` rather than `bundle exec rubocop`

Hope that helps explain it.

rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
```diff
@@ -14,7 +14,10 @@ def enforce_violation(node)
...
+        assignments = rubocop_version.to_f >= 0.59 ? AST::Node::ASSIGNMENTS : Util::ASGN_NODES
```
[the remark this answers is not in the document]
robotpistol: this will not work correctly for semantic versioning.  use Gem::Version (see my comment above)

rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
robotpistol: doh yeah this is a rubocop version check, yeah same principles apply though. just throw in your strinafied version or just use the raw version and compare it to the `Gem::Version.new('2.6.0')` or whatever.

mcamara: Maybe I didn't get your solution but checking for the ruby version will fail if rubocop is updated but ruby isn't. `AST::Node::ASSIGNMENTS` was added in rubocop 0.59 which works for ruby > 2.2

mcamara:
I've updated the rubocop version in the gem to the latest available. 
I've also solved cops that broke the execution of rubocop, but didn't have time  for the unrecognised ones.

mcamara:
Rubocop version updated to the most recent one, rails and performance packages from rubocop added to the gemspec (they were splited from the original rubocop gem.

I've also updated all cops that have been updated since the last version of airbnb/ruby 😄

zachsabin:
I believe the newest version of the master branch (versions >=`3.0.0`) now support the latest rubocop and ruby 2.6. We've also added some new rules.

BMorearty:
Thank you for kicking this off, @mcamara! Much appreciated. I'm going to close this now that we have caught up with Ruby 2.6 on master.

PR #154, opened 2019-02-25
Update to Rubocop 0.60.0

cfrieden:


ljharb:
I think #150 is worth a look.

PR #156, opened 2019-03-26
Update Dependant RuboCop version to allow for newer versions of Ruby

Andy-Bell:
Creates an assignment variable, to allow checking of which method of assignment to use based on which version of rubocop is running (v.0.59.0 introduced the RuboCop::AST::Node::ASSIGNMENTS over the older RuboCop::Util::ASGN_NODES). This should preserve backwards compatibility while allowing for the newer rubocop versions which support v2.6 ruby.

This follows from the commentary on #150 (making use of the idea suggested in the comments by @robotpistol ).

### DIFF
--- a/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
+++ b/rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
@@ -14,7 +14,13 @@ module RuboCop
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        Util::ASGN_NODES.each do |type|
+        assignments =
+          if Gem.loaded_specs['rubocop'].version < Gem::Version.new('0.59.0')
+            RuboCop::Util::ASGN_NODES
+          else
+            RuboCop::AST::Node::ASSIGNMENTS
+          end
+        assignments.each do |type|
           define_method("on_#{type}") do |node|
             enforce_violation(node)
           end
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 0.58.0')
-  spec.add_dependency('rubocop-rspec', '~> 1.30.0')
-  spec.add_development_dependency('rspec', '~> 3.5')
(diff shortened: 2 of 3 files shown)


ljharb:
Thanks! Could we also add newer ruby versions to travis.yml?

rubocop-airbnb/rubocop-airbnb.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 0.58.0')
-  spec.add_dependency('rubocop-rspec', '~> 1.30.0')
-  spec.add_development_dependency('rspec', '~> 3.5')
+  spec.add_dependency('rubocop', '~> 0.62.0')
```
robotpistol: The problem with doing this is that rules update as well. If you are 100% fine with the default provided by rubocop great, but otherwise this will start causing warnings of rules that are misnamed or misconfigured.

rubocop-airbnb/lib/rubocop/cop/airbnb/continuation_slash.rb
```diff
@@ -14,7 +14,13 @@ def enforce_violation(node)
         alias on_send enforce_violation
         alias on_if enforce_violation
 
-        Util::ASGN_NODES.each do |type|
+        assignments =
```
robotpistol: This does look a lot better :)

rubocop-airbnb/rubocop-airbnb.gemspec
ljharb: Does this mean we'd have to have all the core rule names guarded behind a version check?

ghiculescu: do you know if these changes will work with the latest rubocop (0.69 - https://github.com/rubocop-hq/rubocop/commit/c2047b71a68d079533ebe9b3766a5ed80b7f8a42) ?

alejandrok5:
Are you guys planning to merge this any time soon?

BMorearty:
No promises and no timelime, but this week in spare time I started looking at every change in the Robucop changelog between 0.58.0 and 0.69.0 to see how they will affect us.

hans-lizihan:
any updates on this?

BMorearty:
No updates, sorry.

zachsabin:
We've updated to rubocop 0.76 and corrected / modified rules as necessary.

PR #158, opened 2019-07-16
Allow 'indented_relative_to_receiver' option

mattr:
Adds `indented_relative_to_receiver` as a valid option for `Layout/MultilineMethodCallIndentation`.

ref: https://github.com/rubocop-hq/rubocop/issues/1646


pdubb:
Can this be resolved? I'd like to be able to use the indented_relative_to_receiver option, and it seems that overriding the Cop in rubocop.yml still doesn't work.

shashankjo (approved):

Issue #159, opened 2019-07-26
Allowing symbols when specifying class for Airbnb/ClassName and Airbnb/FactoryClassUseString


tomprats:
`Airbnb/ClassName` and `Airbnb/FactoryClassUseString` both work on the idea of not loading the classes. Using a symbol is also a valid way to specify the classes.

```
# Bad
class: MyClass

# Good
class: "MyClass"

# Also Good
class: :MyClass
```

Looking into it, it'd be a pretty simple to add the symbol check. I'd be interested in making the PR if this is a use-case you'll allow

tomprats:
Closing based on the feedback on the corresponding PR https://github.com/airbnb/ruby/pull/160

Perhaps after upgrading to a higher ruby version it'd be more natural, but for now strings are pretty standard

PR #160, opened 2019-07-26
Add symbols as valid class names

tomprats:
Allows symbols to be used as class names

https://github.com/airbnb/ruby/issues/159

### DIFF
--- a/rubocop-airbnb/lib/rubocop/cop/airbnb/class_name.rb
+++ b/rubocop-airbnb/lib/rubocop/cop/airbnb/class_name.rb
@@ -9,7 +9,7 @@ module RuboCop
           'reloading of zeus after changing a model.'.freeze
 
         # Is this a has_many, has_one, or belongs_to with a :class_name arg? Make sure the
-        # class name is a hardcoded string. If not, add an offense and return true.
+        # class name is a hardcoded string or symbol. If not, add an offense and return true.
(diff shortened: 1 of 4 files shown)


tomprats:
@ljharb Was hoping to get your thoughts on this when you get a chance

ljharb:
@tomprats it’d need the thoughts of a current airbnb employee on whether this is a style they want to allow.

BMorearty:
@tomprats Thanks for the PR. What's the rationale for the change?

My inclination is to stick with just strings for consistency but I'd like to hear your reasoning first.

tomprats:
@BMorearty I originally had been using the class until I found out why that was bad. When I was going through fixing things, it just seemed more natural to do `class_name: :Taco` than to do `class_name: "Taco"`

It seemed similar to how we used to do `"hey" => "you"` but now would do `hey: :you` or `hey: "you"` (depending on what's valid or where it was used). Since in this case both strings and symbols are accepted, I thought the rule could allow them as well

BMorearty:
Yeah, I can understand that reasoning. But since both strings and symbols work and we've already been standardizing on strings, let's stick with just strings. Also in the Rails code I've seen, strings are what's typically used when referencing a class by name.

tomprats:
Sounds good. Thanks for taking a look at it guys!

PR #162, opened 2019-11-08
Run 2.6 rubocop

zachsabin:
This PR splits the rubocop configuration based on which version of rubocop is being used. It uses rubocop version 0.76 for Ruby 2.6 and rubocop version 0.58 for <2.6. It sets the version of rubocop based on which version of ruby is being used. Important commits are:
1. https://github.com/airbnb/ruby/pull/162/commits/849605d9ad5e156336e452700d9061931f3e4729 -- sets up the dependency versioning based on which version of ruby is being used (the yml config changes there eventually end up in the 0.76 folder, but they're not super important)
2. https://github.com/airbnb/ruby/pull/162/commits/e1e86de5549b5594d30c6611c14c606a2128ebe4 -- sets up the configuration such that it pulls `yml` files from the relevant folder for the version of rubocop that it's using. 

Next steps if this approach seems reasonable:
1. Running in 2.6 there are a lot of warnings about unrecognized cops. I'll have to go through them at some point and either remove or update them.
2. Ensure that we have coverage of all the styles we want on 0.76
3. Try to reconcile the `freeze` rule on `version.rb` -- 0.76 complains about the freeze but 0.58 requires it. The redundant freeze rule name changed from 0.58 to 0.76, so if you disable it with a comment then 0.58 complains about a redundant disabling. But the name of the rule that checks for redundant disables also changed! So there is no way to get that line to pass without disabling the redundant freeze rule in 0.76 at the moment.


rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -227,7 +227,7 @@ Layout/FirstMethodParameterLineBreak:
   Enabled: false
 
 # Supports --auto-correct
-Layout/FirstParameterIndentation:
+Layout/IndentFirstArgument:
```
ljharb: changing these rule names seems like a breaking change. why?

zachsabin: I copied the new yml configs from https://github.com/airbnb/ruby/pull/150. If the older ones are forwards compatible, then there's no reason to do that (should have checked!)

zachsabin: They are not forwards compatible:
```
Error: The `Layout/FirstParameterIndentation` cop has been renamed to `Layout/IndentFirstArgument`.
(obsolete configuration found in config/0.76/rubocop-layout.yml, please update it)
The `Layout/IndentArray` cop has been renamed to `Layout/IndentFirstArrayElement`.
(obsolete configuration found in config/0.76/rubocop-layout.yml, please update it)
The `Layout/IndentHash` cop has been renamed to `Layout/IndentFirstHashElement`.
(obsolete configuration found in config/0.76/rubocop-layout.yml, please update it)
```

zachsabin: These two commits make it backwards compatible by splitting what config is used based on ruby version:
https://github.com/airbnb/ruby/pull/162/commits/aaf5dd82f13ca962c7cc306a79ac92e3bd99e844
https://github.com/airbnb/ruby/pull/162/commits/1c0591dad875547c1522902bb4c6a5a68581149e

That said it's a bit of a hack

ljharb:
That means rubocop 2.6 is a semver-major, and should have been 3.0

zachsabin:
Well it's really rubocop 0.58 -> 0.76 but I agree! Unfortunately if we want to run stuff with ruby 2.6 then we have to get to a higher version that 0.58.

zachsabin:
Woohoo CI is passing! I will write a longer description of what I did in the PR description. I'm not 100% sure if it's something we want to do but I would really like to be able to run rubocop with 2.6.

rubocop-airbnb/.rubocop.yml
```diff
@@ -1,6 +1,5 @@
 inherit_from:
   - .rubocop_airbnb.yml
-  - ./config/default.yml
```
zachsabin: This did not contribute any configs that weren't already being pulled in by our injection code. I checked via  `bundle exec rubocop --config .rubocop.yml --show-cops` and the same number of cops were being used with and without it.

zachsabin:
Closing as I think we're just going to drop support for ruby versions <2.3, rendering this needlessly complex.

PR #163, opened 2019-11-09
Update to rubocop 0.76

zachsabin:
This will drop support for ruby 2.2 and add it for ruby 2.6. It's probably best reviewed commit by commit as I tried to bundle changes logically -- many cops switched files and that will be more clear if you look solely at commits. 

I looked through the changelogs for rubocop up to 0.76.0 and the changelogs for [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance/blob/master/CHANGELOG.md) and [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails/blob/master/CHANGELOG.md).


rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -474,11 +472,6 @@ Layout/SpaceInsideBlockBraces:
   EnforcedStyleForEmptyBraces: no_space
   SpaceBeforeBlockParameters: true
 
-Layout/SpaceInsideParens:
```
zachsabin: This was just duplicated elsewhere in the file! It still exists, but the newer rubocop doesn't allow duplicated styles it seems.

rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -344,7 +343,6 @@ Layout/MultilineMethodCallIndentation:
   SupportedStyles:
   - aligned
   - indented
-  IndentationWidth:
```
zachsabin: `IndentationWidth` isn't supported anymore, although I can't find in the rubocop changelog where exactly that happened.

rubocop-airbnb/config/rubocop-style.yml
```diff
@@ -360,10 +360,7 @@ Style/FrozenStringLiteralComment:
   Description: Add the frozen_string_literal comment to the top of files to help transition
     from Ruby 2.3.0 to Ruby 3.0.
   Enabled: false
-  EnforcedStyle: when_needed
-  SupportedStyles:
-  - when_needed
-  - always
+  EnforcedStyle: always
```
zachsabin: `when_needed` does not appear to be allowed anymore.
```
Error: invalid EnforcedStyle 'when_needed' for Style/FrozenStringLiteralComment found in config/rubocop-style.yml
Valid choices are: always, never
```

rubocop-airbnb/config/rubocop-style.yml
```diff
@@ -328,11 +328,6 @@ Style/ExpandPathArguments:
   Description: "Use `expand_path(__dir__)` instead of `expand_path('..', __FILE__)`."
   Enabled: false
 
-Style/FlipFlop:
```
zachsabin: Moved to `rubocop-airbnb/config/rubocop-lint.yml`

rubocop-airbnb/config/rubocop-gemspec.yml
```diff
@@ -7,3 +7,6 @@ Gemspec/RequiredRubyVersion:
   Enabled: false
   Include:
     - '**/*.gemspec'
+
+Gemspec/RubyVersionGlobalsUsage:
```
zachsabin: This is enabled by default -- is it worth explicitly enabling here or should we just use the rubocop default settings?

rubocop-airbnb/config/rubocop-layout.yml
BMorearty: The name of the cop is `Layout/IndentationWidth`. This line with bare `IndentationWidth` should probably never have been here, so you were right to remove it.

rubocop-airbnb/config/rubocop-style.yml
```diff
@@ -360,10 +360,7 @@ Style/FrozenStringLiteralComment:
   Description: Add the frozen_string_literal comment to the top of files to help transition
     from Ruby 2.3.0 to Ruby 3.0.
   Enabled: false
-  EnforcedStyle: when_needed
-  SupportedStyles:
-  - when_needed
-  - always
```
BMorearty: Should this stay here, but with `- never` and `- always` as the choices?

rubocop-airbnb/config/rubocop-airbnb.yml
```diff
@@ -2,11 +2,6 @@
 # They are custom built for use inside Airbnb and address issues that we have experienced in
 # testing and production.
 
-Airbnb/ClassName:
```
BMorearty: We can probably delete class_name.rb and class_name_spec.rb now that we're using Rubocop's `ReflectionClassName` cop instead.

rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -551,3 +551,6 @@ Layout/TrailingWhitespace:
   Description: Avoid trailing whitespace.
   StyleGuide: https://github.com/rubocop-hq/ruby-style-guide#no-trailing-whitespace
   Enabled: true
+
+Layout/IndentFirstParameter:
```
BMorearty: It looks like the documented name is `Layout/FirstParameterIndentation`. Let's use that name instead of the alias to make it easier to find documentation.

https://github.com/rubocop-hq/rubocop/blob/a29694f7821ef51ca58528ca703bce9ec713716b/lib/rubocop/config_obsoletion.rb#L17

rubocop-airbnb/config/rubocop-gemspec.yml
BMorearty: Let's explicitly enable it here for clarity. I don't know where the Rubocop defaults are set but I see tons of enabled cops in our configs that are probably also enabled by default.

rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -227,7 +227,7 @@ Layout/FirstMethodParameterLineBreak:
   Enabled: false
 
 # Supports --auto-correct
-Layout/FirstParameterIndentation:
+Layout/IndentFirstArgument:
```
BMorearty: The new name is `Layout/FirstArgumentIndentation`.

rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -238,7 +238,7 @@ Layout/FirstParameterIndentation:
   - special_for_inner_method_call_in_parentheses
 
 # Supports --auto-correct
-Layout/IndentArray:
+Layout/IndentFirstArrayElement:
```
BMorearty: The new name is `Layout/FirstArrayElementIndentation`.

rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -248,10 +248,9 @@ Layout/IndentAssignment:
   Description: Checks the indentation of the first line of the right-hand-side of a
     multi-line assignment.
   Enabled: true
-  IndentationWidth:
 
 # Supports --auto-correct
-Layout/IndentHash:
+Layout/IndentFirstHashElement:
```
BMorearty: The new name is `Layout/FirstHashElementIndentation`

BMorearty:
This is great. I would also like to see this change:

* Disable `Layout/EmptyLineAfterGuardClause` (new cop in 0.58.3).

rubocop-airbnb/config/rubocop-layout.yml
zachsabin: I think all of these `Indent` -> `Indentation` name changes are in rubocop 0.77, which isn't released yet.

https://github.com/rubocop-hq/rubocop/issues/7077

PR #166, opened 2019-11-15
Require gems in library code

zachsabin:
If you depend on this configuration through ruby code rather than through rubocop `.yml` dependencies this change means you don't need to explicitly pull in the `rubocop-performance` and `rubocop-rails` gems in your `yml` files.

### DIFF
--- a/rubocop-airbnb/lib/rubocop-airbnb.rb
+++ b/rubocop-airbnb/lib/rubocop-airbnb.rb
@@ -4,6 +4,9 @@ require 'yaml'
 # Load original rubocop gem
 require 'rubocop'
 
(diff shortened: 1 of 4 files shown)


rubocop-airbnb/lib/rubocop-airbnb.rb
```diff
@@ -4,6 +4,9 @@
 # Load original rubocop gem
 require 'rubocop'
 
+require 'rubocop-performance'
+require 'rubocop-rails'
```
Umofomia: With this addition, shouldn't it mean that this commit is no longer necessary? 72066c174f4f5c3750c51bebf711ca831df5b83a

PR #167, opened 2019-11-15
Fix typo in CHANGELOG

Umofomia:
Fixes minor typo I just noticed.

### DIFF
--- a/rubocop-airbnb/CHANGELOG.md
+++ b/rubocop-airbnb/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 3.02
+# 3.0.2
 * Moves `require`s for `rubocop-performance` and `rubocop-rails` to library code for better transitivity.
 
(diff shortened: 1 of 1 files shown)


rubocop-airbnb/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# 3.02
+# 3.0.2
 * Moves `require`s for `rubocop-performance` and `rubocop-rails` to library code for better transitivity.
```
ljharb: fwiw this seems like a minor, not a patch, but ¯\\\_(ツ)_/¯

Umofomia: It doesn't change how the gem fundamentally functions, but removes the need for consumers to explicitly specify the `require`s. You could view that as a bug fix that merits a patch version bump.

PR #168, opened 2019-12-06
Disable rescue alignment cop until bugs are fixed

zachsabin:
Currently the rescue alignment cop has an issue on variable assignments where it enforces this style:
```
return_val = begin
  # computation
             rescue
               # rescued value
end
```
which is bad. 

https://github.com/rubocop-hq/rubocop/pull/7531 will maybe fix it, but I think it makes sense to disable this for now until it's fixed in a rubocop release.

### DIFF
--- a/rubocop-airbnb/CHANGELOG.md
+++ b/rubocop-airbnb/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 3.1.0
+* Disable Layout/RescueEnsureAlignment
+
 # 3.0.2
 * Moves `require`s for `rubocop-performance` and `rubocop-rails` to library code for better transitivity.
 
--- a/rubocop-airbnb/config/rubocop-layout.yml
+++ b/rubocop-airbnb/config/rubocop-layout.yml
@@ -363,7 +363,7 @@ Layout/MultilineOperationIndentation:
 # Supports --auto-correct
 Layout/RescueEnsureAlignment:
   Description: Align rescues and ensures correctly.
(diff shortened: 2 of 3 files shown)


robotpistol:
The  solution  here is to do this instead
return_val = 
  begin
    # computation
  rescue
    # rescued value
  end

robotpistol:
The problem is more being caused by inconsistent lining up of en with begin block and the body. This isn't a bug but rather configuration

zachsabin:
True, but we'd like the option to have `rescue` align with the variable assignment. It seems like that configuration is in the works, in the meantime the auto correct (what I posted in the PR body) is a very strange format that we'd like to avoid.

zachsabin:
I _think _ the autocorrect bug is now fixed.

I'm going to hold off on merging this for a little bit to see if the rubocop configuration change gets merged. If it seems like it will take a while, then I'll merge this to disable the cop.

michael-zoller-airbnb:
the bug has been [fixed ](https://github.com/rubocop/rubocop/issues/6918#issuecomment-699469222)as of rubocop version [0.9.1](https://github.com/rubocop/rubocop/blob/709c45284aa78ee3c565373b57ec3ac123256c0e/config/default.yml#L376) which add the BeginEndAlignment cop with the default "start_of_line"

pariser:
Since we've progressed well past rubocop 0.9.1, I'm going to close this out. Thanks everyone for the contributions.

PR #171, opened 2020-05-07
Correct 0.80 cops

zachsabin:
These are some of the new cops for 0.80. Here are my rationale for disabling them:

`Rails/RakeEnvironment` -- this dependency is not needed for all rake tasks, so it should not be enforced for all of them.

`Style/HashEachMethods` -- this has an unsafe autocorrect which seems dangerous.

`Style/HashTransformValues` and `Style/HashTransformKeys` should only be used with ruby >= 2.5, which is undesirable.

### DIFF
--- a/rubocop-airbnb/config/rubocop-layout.yml
+++ b/rubocop-airbnb/config/rubocop-layout.yml
@@ -554,3 +554,8 @@ Layout/TrailingWhitespace:
 
 Layout/FirstParameterIndentation:
   Enabled: true
+
+# Supports --auto-correct
+Layout/LineLength:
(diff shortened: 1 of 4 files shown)


rubocop-airbnb/config/rubocop-layout.yml
```diff
@@ -554,3 +554,8 @@ Layout/TrailingWhitespace:
 
 Layout/FirstParameterIndentation:
   Enabled: true
+
+# Supports --auto-correct
```
zachsabin, line 558: This cop now supports auto correct interestingly

PR #178, opened 2020-12-02
Upgrade to latest rubocop packages

bmulholland:
I know this is perhaps not a priority for Airbnb, and likely a project in and of itself to adopt internally, but the cops in this project are super useful for some of us externally. Unfortunately for us non-Airbnbs, the super-tight and now a little out of date packages mean that it's difficult to use alongside other packages, and [especially tricky for adopting in CodeClimate](https://github.com/codeclimate/codeclimate-rubocop/issues/131). It makes it [hard to use this alongside e.g. rubocop-sorbet](https://github.com/codeclimate/codeclimate-rubocop/issues/264).

So I upgraded all the packages to the latest, and allowed any minor version to prevent things from getting so out of date again. The new rspec format took some time, but is IMO much cleaner and more specific. Do note that there's a way to truncate the strings in the expectation blocks by cutting off the /end/, but the `[...]` syntax doesn't work in the beginning or middle of a message. For that reason I've had to include the full text of the message if there's something at the end of the cop output that we want to check.

Even if Airbnb doesn't adopt the latest version of the cops, would you be open to accepting this PR and releasing a new gem version? Alternatively, I suppose I could fork it and issue my own gem, but I don't want to bear /all/ the maintenance burden :)

And either way, check out the one TODO I marked in a spec where the spec didn't match the tested behaviour - what's actually intended?

### DIFF
--- a/rubocop-airbnb/config/rubocop-lint.yml
+++ b/rubocop-airbnb/config/rubocop-lint.yml
@@ -68,10 +68,6 @@ Lint/EmptyInterpolation:
 Lint/EmptyWhen:
   Enabled: false
 
-Lint/EndInMethod:
-  Description: END blocks should not be placed inside method definitions.
-  Enabled: false
-
 Lint/EnsureReturn:
   Description: Do not use return in an ensure block.
   StyleGuide: https://github.com/rubocop-hq/ruby-style-guide#no-return-ensure
@@ -289,8 +285,8 @@ Lint/UselessAssignment:
   StyleGuide: https://github.com/rubocop-hq/ruby-style-guide#underscore-unused-vars
   Enabled: true
 
-Lint/UselessComparison:
-  Description: Checks for comparison of something with itself.
+Lint/BinaryOperatorWithIdenticalOperands:
+  Description: Checks for places where binary operator has identical operands
   Enabled: true
 
 Lint/UselessElseWithoutRescue:
--- a/rubocop-airbnb/config/rubocop-rspec.yml
+++ b/rubocop-airbnb/config/rubocop-rspec.yml
@@ -56,7 +56,6 @@ RSpec/DescribedClass:
 RSpec/EmptyExampleGroup:
   Description: Checks if an example group does not include any tests.
   Enabled: true
-  CustomIncludeMethods: []
 
 RSpec/EmptyLineAfterExampleGroup:
   Description: Checks if there is an empty line after example group blocks.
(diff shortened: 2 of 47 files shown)


bmulholland:
Travis will fail because this change drops support for ruby 2.3

rubocop-airbnb/config/rubocop-lint.yml
```diff
@@ -68,10 +68,6 @@ Lint/EmptyInterpolation:
 Lint/EmptyWhen:
   Enabled: false
 
-Lint/EndInMethod:
```
bmulholland, line 71: Rubocop removed this cop

rubocop-airbnb/config/rubocop-rspec.yml
```diff
@@ -155,10 +154,6 @@ RSpec/InstanceVariable:
   Description: 'Checks for the usage of instance variables.'
   Enabled: false
 
-RSpec/InvalidPredicateMatcher:
```
bmulholland, line 158: Rubocop removed this cop

rubocop-airbnb/spec/rubocop/cop/airbnb/spec_constant_assignment_spec.rb
```diff
@@ -1,80 +1,64 @@
...
+    RUBY
   end
 
-  it 'allows constant defined in global space' do
-    source = [
-      'CONSTANT = 5',
-    ].join("\n")
-
-    inspect_source(source)
-    expect(cop.offenses.size).to eq(1)
```
bmulholland, line 32: FLAG: This test verifies that an error was raised, but the spec text says "allows".

pariser:
@bmulholland thanks for contributing. I just started working on #181 which unfortunately duplicates _some_ of your work here. I wasn't comfortable jumping all the way to rubocop 1.x in one go, so I upgraded to 0.93.1 (the last non-1.x rubocop available). Perhaps we can do this in a two-phase update to minimize the required changes for consumers of this gem?

pariser:
@bmulholland this has been incorporated into https://github.com/airbnb/ruby/pull/189. This, and a few other changes, have been incorporated in our newest [v5.0.0 gem release](https://rubygems.org/gems/rubocop-airbnb/versions/5.0.0) which is now available on rubygems.

Thank you for your contribution, and apologies for the long delay.

cc also @maknz @kelvinlzhang @blueberrystream @bj-mcduck @kkohrt who left emoji reactions on the PR and thus presumably may still be interested in seeing this get over the finish line

PR #193, opened 2023-02-16
Loose rubocop version constraint

RenzoMinelli:
### Why
-------------------
* Currently there's no way to run Rubocop's static code analysis with a target version below 2.5. This was fixed in Rubocop [version 1.30.0](https://github.com/rubocop/rubocop/blob/v1.32.0/CHANGELOG.md#bug-fixes-5). 
* As support for Ruby 2.5 was dropped in [1.29.0](https://github.com/rubocop/rubocop/blob/v1.32.0/CHANGELOG.md#changes-4), we cannot just bump Rubocop version, as it would be a breaking change.
* As `Lint/UselessElseWithoutRescue` was removed in 1.29.0 and added back in  1.31.0, we need to explicitly exclude those versions as we are making use of that Cop.

Therefore in order to support static code analysis and maintain support for currently supported Ruby versions, Rubocop version constraint must be loosen to reach 1.31, excluding the versions which are incompatible.

### DIFF
--- a/rubocop-airbnb/lib/rubocop/airbnb/version.rb
+++ b/rubocop-airbnb/lib/rubocop/airbnb/version.rb
@@ -3,6 +3,6 @@
 module RuboCop
   module Airbnb
     # Version information for the the Airbnb RuboCop plugin.
-    VERSION = '5.0.0'
+    VERSION = '5.1.0'
   end
 end
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,7 +25,10 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.22.0')
+  # rubocop:disable Layout/LineLength
+  spec.add_dependency('rubocop', '>= 1.22.0', '< 1.32', '!= 1.29.0', '!= 1.29.1', '!= 1.30.0', '!= 1.30.1')
+  # rubocop:enable Layout/LineLength
+
   spec.add_dependency('rubocop-performance', '~> 1.10.2')
   spec.add_dependency('rubocop-rails', '~> 2.9.1')
   spec.add_dependency('rubocop-rspec', '~> 2.0.0')


pariser:
> As support for Ruby 2.5 was dropped in [1.29.0](https://github.com/rubocop/rubocop/blob/v1.32.0/CHANGELOG.md#changes-4), we cannot just bump Rubocop version, as it would be a breaking change.

Honestly, Ruby 2.5 is well past EOL. I feel like we could just (a) drop support for 2.5, (b) bump the rubocop version, and then (c) I could put out another major release.

RenzoMinelli:
> > As support for Ruby 2.5 was dropped in [1.29.0](https://github.com/rubocop/rubocop/blob/v1.32.0/CHANGELOG.md#changes-4), we cannot just bump Rubocop version, as it would be a breaking change.
> 
> Honestly, Ruby 2.5 is well past EOL. I feel like we could just (a) drop support for 2.5, (b) bump the Rubocop version, and then (c) I could put out another major release.

I feel like that would leave the major release 5.0.0 without an important feature (static analysis for lower Ruby versions), forcing consumers to upgrade to a new major just to recover that. This could potentially mean that consumers of this gem would need to make a new release to upgrade the dependency just to get this fix – this is the case for the gem `rubocop-airbnb-internal`. Furthermore, there aren't really breaking changes in the code that would make the new version backward incompatible, we are just dropping support for Ruby 2.5, so we can see these changes as a minor release.

In case we make a new major version, what do you think of relaxing the Rubocop constraint in the minor version instead of the patch version, in order to prevent stuff like this from happening again? We could look for the higher Rubocop version that doesn't break our code, and use it as the upper bound for the dependency.

pariser:
> forcing consumers to upgrade to a new major just to recover that

Hey @RenzoMinelli I think you make some good points here, but honestly I don't think it's that big of a problem for most users to just skip straight from their current version right past 5.0.0 directly to this hypothetical 6.0.0 release. [From rubygems.org](https://rubygems.org/gems/rubocop-airbnb), there have only been a few hundred downloads of v5 in the last few days since it was released. If you think about the likely distribution of ruby versions used by those at most 400 people, the chance of them relying on well-past-EOL Ruby2.5 is exceedingly low.
Screenshot 2023-02-17 at 2 35 32 PM

> dropping support for Ruby 2.5, so we can see these changes as a minor release

While ruby 2.5 is a minor version of _ruby_, I'd say that dropping support for a ruby minor version constitutes a major braking change to the functionality of this gem. Also, version numbers are inexpensive, so we shouldn't be afraid to use them. I'd rather be protective of our users and produce major version changes pretty much whenever we have non-backwards compatible behavior. Better to warn that your changes might produce bugs loudly than softly (I'd rather not cause some incident somewhere because someone assumed a gem developer wouldn't possibly drop ruby version support in a minor version gem release. We saw some headaches from AWS doing this exact same thing in their Ruby SDK, and it feels like we can learn from that and not put our users at risk.)

> relaxing the Rubocop constraint in the minor version instead of the patch version, in order to prevent stuff like this from happening again

This one is a little tricky. Minor versions of rubocop can add new cops for which our styleguide won't provide an opinion. We should just do a better job of releasing our own minor version bumps that periodically accept newer minor-version rubocop bumps, but we should do the work to update on our end as we evolve our styleguide (I know - easier said than done)

> We could look for the higher Rubocop version that doesn't break our code, and use it as the upper bound for the dependency.

This is an excellent idea!

RenzoMinelli:
Closed in favor of #194

PR #195, opened 2023-02-27
Use `RESTRICT_ON_SEND`

koic:
Follow up https://github.com/rubocop-hq/rubocop/pull/8365.

This PR uses `RESTRICT_ON_SEND` to restrict callbacks `on_send` to specific method names only. https://docs.rubocop.org/rubocop/1.46/development.html#implementation


pariser (approved):
Thank you!

PR #205, opened 2024-07-16
Support for `Ruby 3.3`

Geromendez135:
Add support for `Ruby 3.3`

Found that `rubocop-airbnb` requires `rubocop 1.32.0` which only supports up to `Ruby 3.2`, See [here](https://github.com/rubocop/rubocop/blame/v1.32.0/lib/rubocop/target_ruby.rb).
[This](https://github.com/rubocop/rubocop/commit/e8a997ad3e01c39659aee8c0d16ef4ede8fde966) is the commit that adds support for `3.3` starting from version `1.46.0`.

# Summary
- Update `rubocop` version to support `Ruby 3.3`
- Constrain `rubocop` version to be `>= 1.61.0` 
- Fixed some deprecations
- Added `Ruby 3.3` to the CI matrix

## Test
- Ran tests locally.

### DIFF
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,9 +25,9 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.32.0')
-  spec.add_dependency('rubocop-performance', '~> 1.10.2')
-  spec.add_dependency('rubocop-rails', '~> 2.9.1')
-  spec.add_dependency('rubocop-rspec', '~> 2.0.0')
+  spec.add_dependency('rubocop', '~> 1.48.0')
+  spec.add_dependency('rubocop-performance', '~> 1.20.2')
+  spec.add_dependency('rubocop-rails', '~> 2.24.0')
+  spec.add_dependency('rubocop-rspec', '~> 2.26.0')
   spec.add_development_dependency('rspec', '~> 3.5')
 end
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.48.0')
+  spec.add_dependency('rubocop', '~> 1.61.0')
   spec.add_dependency('rubocop-performance', '~> 1.20.2')
   spec.add_dependency('rubocop-rails', '~> 2.24.0')
   spec.add_dependency('rubocop-rspec', '~> 2.26.0')
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,9 +25,9 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.61.0')
-  spec.add_dependency('rubocop-performance', '~> 1.20.2')
-  spec.add_dependency('rubocop-rails', '~> 2.24.0')
-  spec.add_dependency('rubocop-rspec', '~> 2.26.0')
+  spec.add_dependency('rubocop', '~> 1.61')
+  spec.add_dependency('rubocop-performance', '~> 1.20')
+  spec.add_dependency('rubocop-rails', '~> 2.24')
+  spec.add_dependency('rubocop-rspec', '~> 2.26')
   spec.add_development_dependency('rspec', '~> 3.5')
 end
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '>= 1.61.0')
+  spec.add_dependency('rubocop', '~> 1.61.0')
   spec.add_dependency('rubocop-performance', '~> 1.20.2')
   spec.add_dependency('rubocop-rails', '~> 2.24.0')
   spec.add_dependency('rubocop-rspec', '~> 2.26.0')
--- a/rubocop-airbnb/rubocop-airbnb.gemspec
+++ b/rubocop-airbnb/rubocop-airbnb.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.61.0')
+  spec.add_dependency('rubocop', '>= 1.61.0')
   spec.add_dependency('rubocop-performance', '~> 1.20.2')
   spec.add_dependency('rubocop-rails', '~> 2.24.0')
   spec.add_dependency('rubocop-rspec', '~> 2.26.0')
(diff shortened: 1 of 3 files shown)


Umofomia:
It looks like the new workflow for 3.3 is [failing](https://github.com/airbnb/ruby/actions/runs/9964224276/job/27681784930?pr=205#step:3:9) because the `ruby/setup-ruby` snapshot that's being used doesn't have 3.3 support. You'll need to update the following to a newer snapshot that has the support:

https://github.com/airbnb/ruby/blob/3e843600e4c58698ddb0c43fa4d71a1c9cf65496/.github/workflows/rspec_rubocop.yml#L34

rubocop-airbnb/rubocop-airbnb.gemspec
```diff
@@ -25,9 +25,9 @@ Gem::Specification.new do |spec|
     'Gemfile',
   ]
 
-  spec.add_dependency('rubocop', '~> 1.32.0')
-  spec.add_dependency('rubocop-performance', '~> 1.10.2')
-  spec.add_dependency('rubocop-rails', '~> 2.9.1')
-  spec.add_dependency('rubocop-rspec', '~> 2.0.0')
+  spec.add_dependency('rubocop', '>= 1.61.0')
```
Umofomia, line 28: Use `~>` instead of `>=` to keep the dependency on the same major version. Also, if Ruby 3.3 support was added in `1.46.0`, would it be better for this to be `~> 1.46.0` instead?

Geromendez135: Yeah, `~>` seems better.
With `rubocop-performance ~> 1.20.2` (which is the first version that tests against `ruby 3.3`) we will need at least `rubocop 1.48.1`, the thing with that version is that when running tests, it results in this error: 

```
➜  rubocop-airbnb git:(gm--support_ruby_3.3) ✗ be rspec

An error occurred while loading spec_helper.
Failure/Error: require 'rubocop-rspec'

RuboCop::ValidationError:
  Property AutoCorrect of cop FactoryBot/CreateList is supposed to be a boolean and contextual is not.
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:261:in `block in check_cop_config_value'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:254:in `each'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:254:in `check_cop_config_value'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:255:in `block in check_cop_config_value'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:254:in `check_cop_config_value'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_validator.rb:34:in `validate'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config.rb:55:in `check'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config.rb:26:in `create'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_loader.rb:57:in `load_file'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-1.48.1/lib/rubocop/config_loader.rb:150:in `inject_defaults!'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-factory_bot-2.26.0/lib/rubocop-factory_bot.rb:16:in `<top (required)>'
# ./vendor/bundle/ruby/3.3.0+0/gems/rubocop-rspec-2.26.1/lib/rubocop-rspec.rb:8:in `<top (required)>'
# ./lib/rubocop/cop/airbnb/spec_constant_assignment.rb:1:in `<top (required)>'
# ./lib/rubocop/airbnb.rb:4:in `map'
# ./lib/rubocop/airbnb.rb:4:in `<top (required)>'
# ./lib/rubocop-airbnb.rb:10:in `<top (required)>'
# ./spec/spec_helper.rb:10:in `<top (required)>'
```

That stops happening from version, 1.61.0 of rubocop, I mention that in [this commit's message](https://github.com/airbnb/ruby/pull/205/commits/95b7a6fe15f776f9c80019676b5ed87ca56c9576)
Also see: https://github.com/rubocop/rubocop/blob/45ba7790b3c6fc37c0b83693513c4dadd619d1ed/CHANGELOG.md?plain=1#L19
[This](https://github.com/rubocop/rubocop/pull/12657) is the PR that adds the marked change in the previous link. Note that 'AutoCorrect: disabled maintains compatibility with AutoCorrect: false.' so I don't think it would be necessary replacing the `false` values.

Geromendez135:
@Umofomia, just updated the setup-ruby snapshot to `v1` let me know whether I should use a specific commit sha or that's OK.

Umofomia:
> let me know whether I should use a specific commit sha or that's OK.

I'm fine with using `@v1`, though wondering if @pariser would like to comment here since he originally introduced the specific commit. [The ruby/setup-ruby documentation](https://github.com/ruby/setup-ruby/blob/161cd54b698f1fb3ea539faab2e036d409550e3c/README.md#versioning) mentions preferring `@v1` anyway, though I can see how a specific commit can be more deterministic.

Geromendez135:
I see it failed for `2.6` because of [this breaking change on rubocop-performance](https://github.com/rubocop/rubocop-performance/blob/master/CHANGELOG.md?plain=1#L94)
should we remove `2.6` from the tests?

Umofomia:
It looks like the updates to the dependencies now require Ruby 2.7, so the Ruby 2.6 tests are now [failing](https://github.com/airbnb/ruby/actions/runs/10014287328/job/27683890499?pr=205#step:3:43). We can probably drop support for Ruby 2.6 anyway going forward, so go ahead and remove the 2.6 tests from the matrix:

https://github.com/airbnb/ruby/blob/bdfad197c3aa6c5ef12b05540868ecdd2ebcc580/.github/workflows/rspec_rubocop.yml#L21-L22

I would also update the line here to indicate that the gem itself only supports 2.7 and higher:
https://github.com/airbnb/ruby/blob/bdfad197c3aa6c5ef12b05540868ecdd2ebcc580/rubocop-airbnb/rubocop-airbnb.gemspec#L18

rubocop-airbnb/rubocop-airbnb.gemspec
```diff
@@ -25,9 +25,9 @@ Gem::Specification.new do |spec|
...
   ]
 
-  spec.add_dependency('rubocop', '~> 1.32.0')
-  spec.add_dependency('rubocop-performance', '~> 1.10.2')
-  spec.add_dependency('rubocop-rails', '~> 2.9.1')
-  spec.add_dependency('rubocop-rspec', '~> 2.0.0')
+  spec.add_dependency('rubocop', '~> 1.61.0')
+  spec.add_dependency('rubocop-performance', '~> 1.20.2')
+  spec.add_dependency('rubocop-rails', '~> 2.24.0')
+  spec.add_dependency('rubocop-rspec', '~> 2.26.0')
```
Umofomia, line 31: I just realized, we should probably remove specification of the patch versions from each of these dependencies (e.g., `~> 1.61` instead of `1.61.0`), to have more flexibility in allowing other minor versions of each dependency.

Geromendez135: Yeah, sounds good to me, justa added a commit with that change!

Umofomia (approved):
Looks good. Thanks for making the change to add support for Ruby 3.3 to the `rubocop-airbnb` gem.

pariser (approved):
This has:
- dropped support for ruby 2.6
- added support for ruby 3.3

This has not:
- changed any of the rubocop rules being applied

Thank you!

